PR #11, opened 2023-12-28
Add Vote Extension processing logic for side txs

Raneet10:
# Description

This PR enables voting for side txs using [Vote Extensions from CometBFT](https://github.com/cometbft/cometbft/blob/main/spec/abci/abci++_methods.md#extendvote). This would eventually replace sidechannel mechanism, which is used for the same in the present.

Please note that the core V.E processing logic has a few dependencies on yet-to-be implemented modules and hence are commented out for now.

# Changes

# Checklist

## Testing

- [ ] I have added unit tests
- [ ] I have tested this code manually on mumbai


lasarojc (commented):
I've done a first pass mostly overlooking the comments addressed at Informal. I will do another one looking into these.

lasarojc (commented):
> Is it okay to not follow the [convention](https://github.com/0xPolygon/cosmos-sdk/blob/507ec1de34858aed7872e60a096f043a975ab3c9/simapp/abci.go#L39) here 👀 ? 

Sure. They don't follow it themselves.

lasarojc (commented):
I see what you mean. `canAddVE` means that the proposal for the current height, the one being extended, could include the `ExtendedVoteInfo` from the previous height. 
If the inclusion is a possibilty only, then the second `if` should not panic when it doesn't find the special transaction.
If `canAddVE` will be extended to respond true only when the special transaction is expected, then this function LGTM.
A simpre approach might be to prefix the marshalled special transaction with some reserved set of bits and check for their presence instead or in compelement to using `canAddVE`.

glnro (commented):
I highly recommend adding some unit tests for `vote_ext_processor.go`

PR #12, opened 2024-01-03
Add app unit tests

Raneet10:
# Description

This PR adds `app` unit tests. The tests currently fail with `error: account address and pubkey address do not match` error. This should be addressed once `auth` module in [our cosmos sdk fork](https://github.com/0xPolygon/cosmos-sdk) uses the Eth address and pub key types instead of `bech32`. 

# Changes

# Checklist

## Testing

- [x] I have added unit tests
- [x] I have tested this code manually on local environment
- [ ] I have tested this code manually on mumbai


marcello33 (commented):
Besides my (minor) comments, tests are not successful for me. I think you need to replace `cosmos-sdk` with the proper version which has the cometBFT changes ported.

marcello33 (approved):

PR #147, opened 2025-02-12
Milestones via vote extensions

avalkov:
# Description

Proposing and agreeing on milestones via vote extensions.


x/milestone/abci/abci.go
```diff
@@ -0,0 +1,173 @@
...
+	logger.Debug("timeSinceLastMilestone", "timeSinceLastMilestone", timeSinceLastMilestone)
+
+	// TODO: make the timeout limit configurable or use a value that depends on block time
+	if pendingMilestone != nil && milestone != nil && timeSinceLastMilestone < 6*time.Second {
+		propStartBlock = pendingMilestone.BlockNumber + 1
+	} else {
+		if milestone == nil {
+			propStartBlock = 0
+		} else {
+			propStartBlock = milestone.EndBlock + 1
```
avalkov, line 53: If we timeout, we will propose the same block that we have proposed already in pendingMilestone no?

I think when we implement the API in bor to lock and return blocks hashes, this logic could lead to perpetual logic.

I was imagining that if we don't reach agreement on milestone in PreBlocker, we set in state some flag like `SkipMilestoneHeight = req.Height+1` and we unlock all blocks in same time to allow bor to reorg.
And we check if our req.Height is equal to `SkipMilestoneHeight` and skip all logic for milestones in every ABCI handler.

avalkov, line 53: If we timeout, we will propose the same block that we have proposed already in pendingMilestone no?

I think when we implement the API in bor to lock and return blocks hashes, this logic could lead to perpetual logic.

I was imagining that if we don't reach agreement on milestone in PreBlocker, we set in state some flag like `SkipMilestoneHeight = req.Height+1` and we unlock all blocks in same time to allow bor to reorg.
And we check if our req.Height is equal to `SkipMilestoneHeight` and skip all logic for milestones in every ABCI handler.

app/abci.go
```diff
@@ -362,6 +389,70 @@ func (app *HeimdallApp) PreBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlo
...
+		}
+
+		if hasMilestone {
+			lastMilestone, err := app.MilestoneKeeper.GetLastMilestone(ctx)
+			if err != nil {
+				logger.Error("Error occurred while fetching the last milestone", "error", err)
+				return nil, err
+			}
+
+			if lastMilestone.EndBlock+1 == majorityMilestone.BlockNumber {
```
avalkov, line 440: Initially I added `BlockNumber` to milestone proposition but then I removed it because I realized if we implement our logic right, it should not be possible to try to finalize milestone out of order.

x/milestone/abci/abci.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	pendingMilestone := GetPendingMilestoneProposition()
+
+	logger := ctx.Logger()
+
+	timeSinceLastMilestone := time.Duration(0)
+
+	if milestone != nil {
+		timeSinceLastMilestone = time.Since(time.Unix(int64(milestone.Timestamp), 0))
```
avalkov, line 41: I am not sure about using time.Since, I think we can have a case where the check passes on some nodes and not on other. I think better to depend on req.Time, which should be the time set by the block proposer.

app/abci.go
cffls: I added it mostly for debugging purpose. Doesn't hurt to keep this check unless it has significant impact on the performance.

x/milestone/abci/abci.go
cffls: Great point, will check if I can use req time here.

x/milestone/abci/abci.go
cffls: This wasn't created as a response to reorgs in bor. It was to resolve a misalignment in milestone proposal. 
In regards to your concern, yes, we can do so as you suggested and unlock all blocks. But before that, we should think about if we need to lock bor at all.

app/abci.go
avalkov: My concern is to not have bug that we could ignore because of this check. If we keep this can we atleast change the logger.Info in the else to logger.Error with big caps, we do that on other places for edge cases that should not happen.

x/milestone/abci/abci.go
avalkov: But instead of checking for time, isn't it more reliable to save at which height the milestone was proposed/finalized and at what height we are now, so we can for sure know that X number of blocks have passed ? Because time check is a bit iffy, there is no guarantee that 6 seconds are 3 blocks because block time is a bit elastic, we can have multiple rounds per block sometimes.

cffls: Yeah, checking block height might be even better.

app/abci.go
cffls: I will change the log level to warning. Ideally this shouldn't happen.

avalkov: Actually I think it can happen, if lets say at block N+1, 6 blocks have passed and we haven't finalized any new milestone, we will propose new milestone starting from the last finalized one but from block N we have proposed other milestone which is more forward, so if it gets 2/3 of votes in the PreBlocker of N+1 we gonna try to finalize it out of order. So we really need the check I think.

Raneet10: If I understood this particular scenario correctly, at N+2 we'll anyway find  common ancestors with the milestone finalized at N+1 right ? So basically we ignore them and finalize from last milestone block at N+1.

avalkov: If we propose 1, 2, 3 at block N, then we propose 4, 5, 6 at N+1 but we finalize only 1, 2 at N+1. At N+2  we wont be able to finalize 4, 5, 6.

x/milestone/abci/abci.go
```diff
@@ -0,0 +1,240 @@
...
+	milestone, err := milestoneKeeper.GetLastMilestone(ctx)
+	if err != nil && err != types.ErrNoMilestoneFound {
+		return nil, err
+	}
+
+	pendingMilestone := GetPendingMilestoneProposition()
+
+	logger := ctx.Logger()
+
+	lastMilestoneBlockNumber, err := milestoneKeeper.GetMilestoneBlockNumber(ctx)
```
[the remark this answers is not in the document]
avalkov: We set it in the PreBlocker now, with the current design the handler and post handler wont be executed anymore, we just need to review to make sure we dont miss something.

x/milestone/abci/abci.go
```diff
@@ -0,0 +1,240 @@
...
+		prefix := make([][]byte, 0)
+		for i := uint64(0); i < blockHashesCount; i++ {
+			prefix = append(prefix, voteExtension.MilestoneProposition.BlockHashes[i])
+
+			hash := common.BytesToHash(bytes.Join(prefix, []byte{'|'})).String()
+			prefixCopy := make([][]byte, len(prefix))
+			copy(prefixCopy, prefix)
+			hashToProp[hash] = &sidetxs.MilestoneProposition{
+				BlockHashes:      prefixCopy,
+				StartBlockNumber: voteExtension.MilestoneProposition.StartBlockNumber,
```
Raneet10, line 121: Can a malicious vote contain the same set of block hashes as the majority but a different `StartBlockNumber`  ?

avalkov: Thats good catch. We should include the `StartBlockNumber` in the `hash := common.BytesToHash(bytes.Join(prefix, []byte{'|'})).String()` because otherwise we can get a case where sometimes we pick the milestone proposition by the malicious proposer and don't finalize, because his start number is wrong.

avalkov: Fixed it.

Raneet10 (approved):

x/milestone/abci/abci.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+
+	params, err := milestoneKeeper.GetParams(ctx)
+	if err != nil {
+		return err
+	}
+
+	if len(milestoneProp.BlockHashes) > int(params.MaxMilestonePropositionLength) {
+		return fmt.Errorf("too many blocks in proposition")
+	}
```
cffls, line 306: Another thought... If there is a limit on the size of vote extension message, we would already have an implicit limit on the number of blocks a validator can put into a milestone prop. In that case, do we still need this validation?

cffls: If the concern is about the calculation time, we can still use `MaxMilestonePropositionLength` and put a hard stop in `GetMajorityMilestoneProposition`.

marcello33 (approved):

PR #169, opened 2025-02-20
changes to packaging templates

marcello33:
# Description

In this PR, we update the packaging templates based on the new heimdall-v2 configs.

# Changes

# Checklist


packaging/templates/config/amoy/app.toml
```diff
@@ -0,0 +1,289 @@
...
+[api]
+
+# Enable defines if the API server should be enabled.
+enable = true
+
+# Swagger defines if swagger documentation should automatically be registered.
+swagger = false
+
+# Address defines the API server to listen on.
+address = "tcp://localhost:1317"
```
kamuikatsurgi, line 136: We should wait till we confirm this from the informal team.

packaging/templates/config/mainnet/app.toml
```diff
@@ -0,0 +1,289 @@
...
+[api]
+
+# Enable defines if the API server should be enabled.
+enable = true
+
+# Swagger defines if swagger documentation should automatically be registered.
+swagger = false
+
+# Address defines the API server to listen on.
+address = "tcp://localhost:1317"
```
kamuikatsurgi, line 136: Same.

marcello33: Yes, thanks, let's wait ;)

packaging/templates/config/amoy/app.toml
kamuikatsurgi: @marcello33 We can have `"tcp://localhost:1317"` as the default but should we add instructions in the comments just above it on the use cases of using `0.0.0.0`?

marcello33: Did the way around (same as v1), but left a comment
Check this out 1d2d9dc3ac10020f1b57cf9e87cafa10731424d3

packaging/templates/config/mainnet/app.toml
marcello33: Did the way around (same as v1), but left a comment
Check this out 1d2d9dc3ac10020f1b57cf9e87cafa10731424d3

djpolygon (approved):
LGTM, we should test profiles and end to end

avalkov (approved):

PR #177, opened 2025-02-26
get headers via batch rpc

lucca30:
# Description

Batch calls instead of multiple gRPC requests. This provides more performance since replaces 10 multiple calls by a single one. 

# Changes

# Checklist

## Testing

- [ ] I have added unit tests

### DIFF
--- a/helper/call.go
+++ b/helper/call.go
@@ -534,7 +534,6 @@ func (c *ContractCaller) GetBorChainBlock(ctx context.Context, blockNum *big.Int
 
 // GetBorChainBlock returns bor chain block headers via single RPC Batch call
 func (c *ContractCaller) GetBorChainBlocksInBatch(ctx context.Context, start, end int64) ([]*ethTypes.Header, error) {
-	Logger.Info("#################### Validating GetBorChainBlocksInBatch Being called", "start", start, "end", end)
 	ctx, cancel := context.WithTimeout(ctx, c.BorChainTimeout)
 	defer cancel()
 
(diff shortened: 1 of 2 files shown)


helper/call.go
```diff
@@ -531,6 +532,44 @@ func (c *ContractCaller) GetBorChainBlock(ctx context.Context, blockNum *big.Int
...
+
+	// Execute the batch call.
+	if err := rpcClient.BatchCallContext(ctx, batchElems); err != nil {
+		return nil, err
+	}
+
+	// Collect the results.
+	var response []*ethTypes.Header
+	for i, elem := range batchElems {
+		if elem.Error != nil {
```
avalkov, line 563: I know it should not be able to happen but just to protect ourselves from some future problems in bor, can we check if we didn't returned error for middle block. I mean if we get response 1, 2, 3, error, 5, 6, 7, we should return only sequential hashes 1, 2, 3.

helper/call.go
```diff
@@ -531,6 +532,44 @@ func (c *ContractCaller) GetBorChainBlock(ctx context.Context, blockNum *big.Int
...
+			// ignore errors since it can be block not found
+			continue
+		}
+		response = append(response, result[i])
+	}
+
+	return response, nil
```
avalkov, line 570: Is it guaranteed that response here will be sorted in ascending order from start to end found blocks?

lucca30: Yes, in this case will because we actually iterate over the own array we built on the request:

### The construction:
https://github.com/0xPolygon/heimdall-v2/blob/a824257a6794120128e955ce35bfb41901f405b3/helper/call.go#L542-L553

### The request:
https://github.com/0xPolygon/heimdall-v2/blob/a824257a6794120128e955ce35bfb41901f405b3/helper/call.go#L556


### The iteration on response:
https://github.com/0xPolygon/heimdall-v2/blob/a824257a6794120128e955ce35bfb41901f405b3/helper/call.go#L561-L568

helper/call.go
lucca30: Should we really just return 1,2,3 or fail the process? If just 1,2,3 is fine, I can replace `continue` by `break`

avalkov: Lets return just 1, 2, 3.

avalkov (approved):

helper/call.go
lucca30: Ok, done

helper/call.go
```diff
@@ -531,6 +532,43 @@ func (c *ContractCaller) GetBorChainBlock(ctx context.Context, blockNum *big.Int
...
+		return nil, err
+	}
+
+	// Collect the results.
+	var response []*ethTypes.Header
+	for i, elem := range batchElems {
+		if elem.Error != nil {
+			break
+		}
+		response = append(response, result[i])
```
cffls, line 566: It might be a good idea to check if the result is nil and break the loop if it is, e.g. when a block number doesn't exist.

lucca30: You have a point! 
I was assuming `elem.Error != nil` will catch this scenario. But probably it won't. I'm deploying a node here to check.

lucca30: Check included:

https://github.com/0xPolygon/heimdall-v2/blob/344361ed44dcca4064baf6e145f2e4f379bb54c1/helper/call.go#L563

cffls (approved):

PR #186, opened 2025-03-05
Initial optimizations on PrepareProposal/ProcessProposal for EL block production with Engine API 

Raneet10:
# Description

This PR is a first attempt to pipeline the EL payload production via the ABCI stages (Prepare/ProcessProposal mostly). The results saw a good improvement in `PrepareProposal` stage but needs to be tested for various other corner cases. 

Before
```
12:20PM INF 🕒 End PrepareProposal: duration=88.973923ms height=30 module=server momentTime=20:34.164467 payloadSize=296102
```

After
```
12:08PM INF 🕒 End PrepareProposal: duration=1.594101ms height=53 module=server momentTime=08:50.719651 payloadSize=292562
```

# Changes

# Checklist

## Testing

- [ ] I have added unit tests
- [x] I have tested this code manually on local environment


app/abci.go
```diff
@@ -236,23 +251,57 @@ func (app *HeimdallApp) NewProcessProposalHandler() sdk.ProcessProposalHandler {
...
 			return &abci.ResponseProcessProposal{Status: abci.ResponseProcessProposal_REJECT}, nil
 		}
 
+		// app.execPayloadCache.Add(payload.LatestValidHash, payload)
 		duration := time.Since(start)
 		formatted := fmt.Sprintf("%.6fms", float64(duration)/float64(time.Millisecond))
-		logger.Info("🕒 End ProcessProposal:", "duration", formatted, "height", req.Height, "payloadSize", len(req.Txs[0]), "momentTime", time.Now().Format("04:05.000000"))
+		logger.Info("🕒 ProcessProposal:" + formatted)
+
+		app.nextBlockChan <- nextELBlockCtx{height: req.Height + 1,
```
avalkov, line 297: Where is `nextExecPayload` read from?

app/abci.go
```diff
@@ -236,23 +251,57 @@ func (app *HeimdallApp) NewProcessProposalHandler() sdk.ProcessProposalHandler {
 		var executionPayload engine.ExecutionPayload
 		err = json.Unmarshal(metadata.MarshaledExecutionPayload, &executionPayload)
 		if err != nil {
+			app.currBlockChan <- nextELBlockCtx{height: req.Height,
```
avalkov, line 254: Why we don't simply return `app.currBlockChan <- nextELBlockCtx{height: req.Height, context: ctx}` in these cases ?

app/abci.go
```diff
@@ -58,51 +58,56 @@ func (app *HeimdallApp) NewPrepareProposalHandler() sdk.PrepareProposalHandler {
 		}
 
 		// Engine API
-		latestBlock, err := app.caller.BorChainClient.BlockByNumber(ctx, big.NewInt(req.Height-1)) // change this to a keeper
-		if err != nil {
-			return nil, err
-		}
+		var payload *engine.Payload
+		if app.latestExecPayload != nil {
+			payload = app.latestExecPayload
```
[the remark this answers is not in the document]
Raneet10: Yes good point. Another option is to set `app.latestExecPayload` to nil once we use it but the height check approach seems better.

app/abci.go
Raneet10: Saves a bor RPC call.

avalkov: What will be consequences if the ParentHash is invalid? Proposed by malicious or corrupted proposer.

Raneet10: Actually we can use the last block value stored in the keeper.

avalkov: Ideally thats what we should use, instead of fetching latest block number from erigon. Fetching of the latest block number from erigon was just a temporary workaround.

PR #196, opened 2025-03-13
Milestones fast forward

avalkov:
# Description

Added check to verify that blocks we gonna propose for milestone are on the same fork as the last finalized milestone.
Added logic to finalize bigger range of blocks if difference between last finalized milestone and execution layer tip is too big.


x/milestone/abci/abci.go
```diff
@@ -276,14 +295,30 @@ func GetMajorityMilestoneProposition(validatorSet *stakeTypes.ValidatorSet, extV
...
 	result := make([][]byte, 0)
 
 	headers, err := contractCaller.GetBorChainBlocksInBatch(ctx, int64(startBlock), int64(startBlock+maxBlocksInProposition-1))
 	if err != nil {
 		return nil, fmt.Errorf("failed to get headers")
 	}
 
+	if len(headers) > 0 && len(lastMilestoneHash) > 0 {
+		lastFinalizedFork := headers[0].ParentHash.Bytes()
+		if startBlock-lastMilestoneBlock > 1 {
```
lucca30, line 308: To make it more readable. What about: `if startBlock >lastMilestoneBlock`

x/milestone/abci/abci.go
```diff
@@ -276,14 +295,30 @@ func GetMajorityMilestoneProposition(validatorSet *stakeTypes.ValidatorSet, extV
...
+			headers, err := contractCaller.GetBorChainBlocksInBatch(ctx, int64(lastMilestoneBlock), 1)
```
lucca30, line 309: This will not get any block unless `int64(lastMilestoneBlock) <=1`. The method expects first the `start` param and after the `end`

```
https://github.com/0xPolygon/heimdall-v2/blob/74329a0ca00e7b503d08a11af605648a0970da10/helper/call.go#L546-L555
```

Also, when I implemented this method I was assuming we were going to call it on small ranges. If this range can potentially be big, we need to improve the `GetBorChainBlocksInBatch` method. Because bor clients can limit the amount of items in batch. In general this number is big, but some RPC providers, for example, put this as 10. So we maybe need to introduce a config variable `maxBorBatch` that limits the batches calls.

x/milestone/abci/abci.go
avalkov: Here I am checking if we are in the case where next milestone start block will be right after last finalized milestone or it will be 1000 blocks in the future lets say.

lucca30: Oh, my bad. Yeah, now i realised the point. I don't know why I looked to this like `if startBlock-lastMilestoneBlock >= 1` or `if startBlock-lastMilestoneBlock > 0`. I see your point

x/milestone/abci/abci.go
avalkov: Thanks, replaced it with call to get single header.
We wont call it for big ranges, we just need to get the current execution client fork block right after the last finalized block number and check via parent relationship if they are same fork.

lucca30: Nice. The best scenario possible

lucca30 (approved):

x/milestone/abci/abci.go
```diff
@@ -166,11 +196,39 @@ func GetMajorityMilestoneProposition(validatorSet *stakeTypes.ValidatorSet, extV
...
+		key := getParentChildKey(parentHash, common.BytesToHash(blockToHash[majorityBlocks[0]]).String())
+		if parentHashToVotingPower[key] >= majorityVP {
+			isParentHashMajority = true
+			majorityParentHash = parentHash
+			break
+		}
+	}
+
+	if !isParentHashMajority {
+		logger.Debug("No parent hash found with majority support")
```
lucca30, line 212: I suppose this log you used for some validation. But still necessary on the application for future debug?

x/milestone/abci/abci.go
```diff
@@ -276,19 +334,36 @@ func GetMajorityMilestoneProposition(validatorSet *stakeTypes.ValidatorSet, extV
 	return proposition, aggregatedProposersHash, supportingValidatorList[0], nil
 }
 
-func getBlockHashes(ctx sdk.Context, startBlock, maxBlocksInProposition uint64, contractCaller helper.IContractCaller) ([][]byte, error) {
+func getBlockHashes(ctx sdk.Context, startBlock, maxBlocksInProposition uint64, lastMilestoneHash []byte, lastMilestoneBlock uint64, contractCaller helper.IContractCaller) ([]byte, [][]byte, error) {
```
lucca30, line 337: This whole new logic of latest milestone hash check doesn't seems to fit in this function. I'd suggest either rename this function or maybe create a function specifically for this task of checking the first block of the next milestone against the latest milestone hash.

x/milestone/abci/abci.go
```diff
@@ -27,42 +28,60 @@ func GenMilestoneProposition(ctx sdk.Context, milestoneKeeper *keeper.Keeper, co
...
-	propStartBlock := uint64(0)
+		params, err := milestoneKeeper.GetParams(ctx)
+		if err != nil {
+			return nil, err
+		}
 
-	if milestone != nil {
-		propStartBlock = milestone.EndBlock + 1
+		if latestHeader.Number.Uint64() > milestone.EndBlock && latestHeader.Number.Uint64()-milestone.EndBlock > params.FfMilestoneThreshold {
+			propStartBlock = ((latestHeader.Number.Uint64() - milestone.EndBlock) / params.FfMilestoneBlockInterval) * params.FfMilestoneBlockInterval
```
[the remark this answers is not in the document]
avalkov: I added new variable.

Yes, I think your right, there should be +1.

x/milestone/abci/abci.go
avalkov: I didn't used it for validation, we add such logs for incase in future we have to understand why the network doesnt reach finality on milestones. I even would want them to be warning or error level but Jerry suggested just to keep them debug.

x/milestone/abci/abci.go
avalkov: Moved the parent-child relationship check in separate function call.

lucca30 (approved):

cffls (commented):
A minor comment, but looks great overall.

cffls (approved):

PR #202, opened 2025-03-19
Engine module

christophercampbell:
# Description

This PR wires in an initial engine module for keeper storage of execution state.

Ticket: https://polygon.atlassian.net/browse/POS-2888

# Changes

# Checklist

# Cross repository changes

  - In case link the PR here:
  - In case link the PR here:

## Testing

- [ ] I have added unit tests

# Additional comments

This is a work in progress. It currently cannot initialize a chain, as the engine genesis is not showing up in the default genesis


app/app.go
```diff
@@ -64,7 +64,7 @@ import (
 	"github.com/gorilla/mux"
 
 	"github.com/0xPolygon/heimdall-v2/client/docs"
-	"github.com/0xPolygon/heimdall-v2/engine"
+	engineclient "github.com/0xPolygon/heimdall-v2/engine" // should this code be moved to module?
```
avalkov, line 67: Maybe yes, you can move it under client package in the module.

avalkov (approved):

avalkov:
I think maybe its good idea to to register it in cosmos sdk `gov` module so its allowed to update params via gov proposal in the future - https://github.com/0xPolygon/cosmos-sdk/blob/devel/x/gov/keeper/msg_validator.go#L33
The params are empty now, but we will need atleast one flag if its enabled or disabled, which will be separate task.

christophercampbell:
Okay that sounds right, but can't add it there until this module is included in a release tag.

PR #216, opened 2025-05-02
Milestones - Migration, Verification, Import, Export

avalkov:
# Description

Milestones - Migration, Verification, Import, Export


cmd/heimdalld/cmd/migration/verify/verify.go
```diff
@@ -873,3 +866,83 @@ type validatorBasicInfo struct {
...
+
+		hashBytes, err := hex.DecodeString(cp["hash"].(string)[2:])
+		if err != nil {
+			return fmt.Errorf("failed to decode root_hash at index %d: %w", i, err)
+		}
+		if !bytes.Equal(milestone.Hash, hashBytes) {
+			return fmt.Errorf("mismatch in milestone root hash at index %d: expected %x, got %x", i, hashBytes, milestone.Hash)
+		}
+	}
+
```
marcello33, line 939: I would also add the following here:

```
		if milestone.MilestoneId != cp["milestone_id"].(string) {
			return fmt.Errorf("mismatch in milestone id at index %d: expected %s, got %s", i, cp["milestone_id"], milestone.MilestoneId)
		}

		// validate that milestoneID is composed by `UUID - HexAddressOfTheProposer`
		splitMilestoneID := strings.Split(milestone.MilestoneId, " - ")
		if len(splitMilestoneID) != 2 {
			return fmt.Errorf("invalid milestoneID %s, it should be composed by `UUID - HexAddressOfTheProposer`", splitMilestoneID)
		}

		_, err = uuid.Parse(splitMilestoneID[0])
		if err != nil {
			return fmt.Errorf("invalid milestoneID, first part is not a valid uuid %s", splitMilestoneID[0])
		}

		if !common.IsHexAddress(splitMilestoneID[1]) {
			return fmt.Errorf("invalid milestoneID, second part is not a valid address %s", splitMilestoneID[1])
		}
		
		if splitMilestoneID[1] != milestone.Proposer {
			return fmt.Errorf("invalid milestoneID, second part is not the proposer address %s", splitMilestoneID[1])
		}

```

WDYT?

cmd/heimdalld/cmd/migration/verify/verify.go
```diff
@@ -873,3 +866,83 @@ type validatorBasicInfo struct {
...
+		return fmt.Errorf("failed to cast checkpoint module data")
+	}
+
+	milestones, ok := checkpointData["milestones"].([]interface{})
+	if !ok {
+		return fmt.Errorf("milestones not found or invalid format")
+	}
+
+	if milestones == nil {
+		milestones = []interface{}{}
```
marcello33, line 888: I think this is not enough for the following case (which I had to fix for `txSequences` in `topup` module.
If we don't have milestones from v1, I would handle both cases in the json file, meaning `milestones = null` and `milestones = []`.

marcello33: (this can be verified by running the `TestRunMigrationVerification` with modified data sets for v1 milestones.

cmd/heimdalld/cmd/migration/verify/verify.go
avalkov: Second part is not the `HexAddressOfTheProposer` but bor block hash https://github.com/maticnetwork/heimdall/blob/master/bridge/setu/processor/milestone.go#L168

Added all except of the last check for the proposer.

cmd/heimdalld/cmd/migration/verify/verify.go
avalkov: Fixed.

cmd/heimdalld/cmd/migration/verify/verify.go
marcello33: Oh yeah, my bad. Thanks

marcello33 (approved):

PR #221, opened 2025-05-07
Total Difficulty on Milestones

lucca30:
# Description

This PR introduces two key changes:

1. **Include Total Difficulty in Milestone Payload**  
   - Adds a `total_difficulty` field to the `MilestoneProposition` and `Milestone` protobuf and its payload.  
   - This is required by stateless clients when processing fast-forwarded blocks, as they need the chain’s cumulative difficulty to validate and sync correctly.

2. **Panic Recovery in `ExtendVoteHandler`**  
   - Wraps the entire handler logic with a `defer`+`recover` block.  
   - On panic, captures and logs the full goroutine stack to aid debugging and pinpoint the exact failure site, rather than crashing uninformatively.

# Changes

- [x] Breaking change (change that is not backwards-compatible and/or changes current functionality)

# Breaking changes

Changes on proto files

# Checklist

- [ ] I have added at least 2 reviewer or the whole pos-v1 team
- [ ] I will be resolving comments - if any - by pushing each fix in a separate commit and linking the commit hash in the comment reply

# Cross repository changes

- [x] This PR requires changes to bor 
  - In case link the PR here: TBD
  - In case link the PR here:


app/abci.go
```diff
@@ -421,21 +435,29 @@ func (app *HeimdallApp) PreBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlo
 
 		addMilestoneCtx, msCache := app.cacheTxContext(ctx)
 
+		td, err := app.caller.GetBorChainBlockTd(ctx, common.BytesToHash(majorityMilestone.BlockHashes[len(majorityMilestone.BlockHashes)-1]))
+		if err != nil {
+			logger.Error("Error occurred while getting td value from last block", "error", err)
+			return nil, err
+		}
+
```
cffls, line 443: Not sure if this is safe. We might need to add `td` in [milestone proposal](https://github.com/0xPolygon/heimdall-v2/blob/develop/x/milestone/abci/abci.go#L71-L75) itself. There are two reasons:
1. Although unlikely, if `td` is different on different nodes, they will end up in different, non-deterministic hash. 
2. In order for heimdall (non-validators) to sync, this change requires bor has caught up and is in sync, which is a strong requirement that we don't enforce before.

lucca30: Thanks for flagging this—I agree with both points.
My original thought was to avoid querying TD on every block to keep the payload light—but since milestones are much shorter now, the extra bytes are really minimal. And I completely agree that making an HTTP call from the preblocker breaks its deterministic, stateless contract.

lucca30: Fixes applied on 65677b51df0cd7d5e4e673910ad905e96a75f455

Now TDs are captured with headers while building proposition. The hash chosen on Majority algorithm comes with the TD capture by the proposer.

I'm just concerned if it's worth also include a comparison between TDs to ensure the agreement. I'm assuming they are equal since it's already agreed the block number and the difficulty (because of hash).

lucca30: Including @marcello33 and @avalkov on discussion.

lucca30: Humm, I see your point. So I'll include the td as part of the key when counting the voting power. In this way we make sure both TD and Hash are agreed for the consensus

lucca30: @avalkov , change applied on fa959efd2afc94a6cef07a878b712829c4a876c2

app/abci.go
```diff
@@ -156,6 +157,19 @@ func (app *HeimdallApp) ExtendVoteHandler() sdk.ExtendVoteHandler {
...
+				buf := make([]byte, 1<<16)
+				n := runtime.Stack(buf, false)
+				logger.Error(
+					"panic in ExtendVoteHandler",
+					"panic", r,
+					"stack", string(buf[:n]),
+				)
+				panic(r)
+			}
+		}()
```
cffls, line 172: Probably safe to remove?

x/milestone/abci/abci.go
```diff
@@ -330,10 +338,21 @@ func GetMajorityMilestoneProposition(
 		)
 	}
 
+	// splitting back blockHashAndTd
+	blockHashes := make([][]byte, 0, len(blockHashesAndTds))
+	blockTds := make([]uint64, 0, len(blockHashesAndTds))
+	for _, blockHashAndTd := range blockHashesAndTds {
+		tdBytes := blockHashAndTd[len(blockHashAndTd)-8:] // last 8 bytes are the TD
```
cffls, line 345: Although we know td will be uint64, can we enforce it to be encoded to 8 bytes [above](https://github.com/0xPolygon/heimdall-v2/pull/221/files#diff-d05469dcea6f221d670bcdcddc8d5b6e5410fad53d8cb84d5ac029f519f01113R160-R162)?

x/milestone/abci/abci.go
```diff
@@ -155,11 +157,17 @@ func GetMajorityMilestoneProposition(
 
 		prop := voteExtension.MilestoneProposition
 		for i, blockHash := range prop.BlockHashes {
+			blockTd := prop.BlockTds[i]
```
avalkov, line 160: We need to add validation inside `ValidateMilestoneProposition` that len(BlockTds) == len(BlockHashes).

x/milestone/abci/abci.go
```diff
@@ -155,11 +157,17 @@ func GetMajorityMilestoneProposition(
...
+			var buf bytes.Buffer
+			binary.Write(&buf, binary.LittleEndian, blockTd)
```
avalkov, line 162: Missing error check.

app/abci.go
lucca30: Yes, it is. But the overhead we gain by including a `defer` I assume it's really small compared by the gain we have on keeping it.

Since the `VoteExtension` works as a goroutine, whenever there's is a panic on it we don't log the line where it failed on the goroutine. With this extra handling actually do it, so i think it's handful in a production scenario.

x/milestone/abci/abci.go
lucca30: enforced on 501ab3e7398a8c191a02f860db9503096b399d3d

avalkov (approved):

app/abci.go
cffls: Sounds good, thanks!

cffls (approved):

PR #283, opened 2025-06-07
Veblop producer selection

cffls:
# Description

The implementation of [validator-elected block producer](https://github.com/maticnetwork/Polygon-Improvement-Proposals/blob/main/PIPs/PIP-64.md) (VeBlop). 
Corresponding PR for bor changes: https://github.com/maticnetwork/bor/pull/1567 

## Major changes
* Introduced a new message type `MsgVoteProducers`, which could be used by validators to cast their votes on producers.
  * A producer vote contains a list of candidate Producers rank by their priorities. The ones whose index is lower in the list will have higher priority.
* Enabled block producer selection based on the producer votes.
  * A fix number of block producers will be selected after filtering, based on aggregated voting power. 
* Removed span proposal and span vote.
* Added the milestone voting performance of each validator.
* Added the ability to automatically create future spans.
* Introduce dynamic span rotation when the current producer fails to create blocks that could be finalized (by milestone) within a time window.

## Migration plan

Unlike past Heimdall hardforks, VeBlop HF will be enabled based on the observation on bor's milestone. 

**Phases**
1. Upgrade bor. The upgrade will specify a block number, `N`, for VeBlop HF, after which each span will contain only one block producer.
2. Upgrade heimdall. The upgrade will use the same block number `N` specify in phase 1. After the upgrade, the new client will be compatible with old clients.
3. One span before VeBlop HF block, each validator will automatically start voting for their preferred block producers. At this point, all validators are supposed to upgrade to the new client. Old client will run into consensus failure.
4. Once enough votes are collected, a new VeBlop span, with start block num `N`, will be proposed and adopted by bor.

```
VeBlop Hardfork Migration Timeline
=====================================

Block Timeline:    ... -----> [N-1 Span] -----> [Block N] -----> [VeBlop Span] -----> ...
                                 |                   |                    |
                                 |                   |                    |
Phase 1: Bor Upgrade             |                   |                    |
┌─────────────────────────────┐  |                   |                    |
│ • Bor clients upgraded      │  |                   |                    |
│ • VeBlop HF block N defined │  |                   |                    |
│ • Ready for single producer │  |                   |                    |
│   spans after block N       │  |                   |                    |
└─────────────────────────────┘  |                   |                    |
                                 |                   |                    |
Phase 2: Heimdall Upgrade        |                   |                    |
┌─────────────────────────────┐  |                   |                    |
│ • Heimdall clients upgraded │  |                   |                    |
│ • Same block number N used  │  |                   |                    |
│ • Backward compatibility    │  |                   |                    |
│   with old clients          │  |                   |                    |
└─────────────────────────────┘  |                   |                    |
                                 |                   |                    |
Phase 3: Voting Phase            |                   |                    |
┌─────────────────────────────┐  |                   |                    |
│ • One span before block N   │<-+                   |                    |
│ • Validators vote for       │                      |                    |
│   preferred producers       │                      |                    |
│ • All validators MUST       │                      |                    |
│   upgrade (old = failure)   │                      |                    |
│ • Vote collection begins    │                      |                    |
└─────────────────────────────┘                      |                    |
                                                     |                    |
Phase 4: VeBlop Activation                           |                    |
┌─────────────────────────────┐                      |                    |
│ • Sufficient votes collected│                      |                    |
│ • New VeBlop span proposed  │                      |                    |
│ • Single producer format    │ ─────────────────────+────────────────────+
│ • Span starts at block N    │                                           |
│ • Bor adopts new span       │ <─────────────────────────────────────────+
└─────────────────────────────┘

Legend:
┌───┐  Phase boundaries
│   │  Phase activities  
 <──  Dependencies/triggers
 ───> Timeline progression

Critical Points:
• Block N: VeBlop hardfork activation block
• N-1 Span: Last multi-producer span, voting begins
• VeBlop Span: First single-producer span
• Old clients become incompatible in Phase 3
``` 

## Breaking changes

Introduces VeBlop hardfork.

## Testing

Added unit tests for major functions.

E2E test in a kurtosis cluster. Setup:
  * Span length = 128
  * 7 validators with veblop enabled (new client)
  * 1 validator without veblop (old client)
  * VeBlop block scheduled at block 383 (start block of span 3)
  * Old client node was able to sync up to block 367, which was one sprint before the beginning of span 3
  * Old span mechanism (span proposal + vote) successfully created span 1 and 2.
  * Validators were able to vote for preferred validator starting in span 2 (256-382)
  * VeBlop span was successfully created for block 383 to 511. 
  * New clients (bor+heimdall) was able to produce/sync blocks without interruption.


app/abci.go
```diff
@@ -341,6 +377,167 @@ func (app *HeimdallApp) VerifyVoteExtensionHandler() sdk.VerifyVoteExtensionHand
...
+		}
+
+		delete(latestActiveProducer, currentProducer)
+
+		err = app.BorKeeper.AddNewVeblopSpan(addSpanCtx, currentProducer, lastMilestone.EndBlock+1, endBlock, lastMilestone.BorChainId, latestActiveProducer)
+		if err != nil {
+			logger.Warn("Error occurred while adding new veblop span", "error", err)
+		} else {
+			// update the last milestone block to a future block height to avoid immediately rotating the span in the next block
+			err = app.MilestoneKeeper.SetLastMilestoneBlock(addSpanCtx, uint64(ctx.BlockHeight())+SpanRotationBuffer)
```
avalkov, line 521: So incase next producer is not ready, we can have ~2 minutes down time before rotating again?

cffls: I was assuming 60 blocks is 1 min. It might be too long, and we can probably reduce it once the point above (span <> milestone conflict) is addressed.

x/bor/keeper/veblop.go
```diff
@@ -0,0 +1,426 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// SelectNextSpanProducer selects the next producer for a new span.
+// It calculates candidate set, filters by active producers and selects one.
+func (k *Keeper) SelectNextSpanProducer(ctx context.Context, currentProducer uint64, activeValidatorIDs map[uint64]struct{}) (uint64, error) {
+	candidates, err := k.CalculateProducerSet(ctx)
```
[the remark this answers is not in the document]
cffls: It is possible when consensus can't be reached. In that case, we need to produce a default list. Will add this.

avalkov (approved):

PR #319, opened 2025-06-26
feat: add self-heal for checkpoints

kamuikatsurgi:


marcello33 (approved):

bridge/listener/rootchain_selfheal.go
```diff
@@ -69,6 +82,99 @@ func (rl *RootChainListener) startSelfHealing(ctx context.Context) {
...
+
+	if l1HeaderBlockId == ackCount {
+		rl.Logger.Info("Latest checkpoint is already synced on Heimdall; skipping", "l1HeaderBlockId", l1HeaderBlockId, "heimdallAckCount", ackCount)
+		return
+	}
+
+	// Check if we have a checkpoint in buffer.
+	bufferedCheckpoint, err := util.GetBufferedCheckpoint(rl.cliCtx.Codec)
+	if err != nil {
+		rl.Logger.Info("No buffered checkpoint found")
```
marcello33, line 127: Is this the case?

marcello33:
I'd like to have this merged too before mainnet release. Please review @avalkov @Raneet10 @pratikspatil024 
Thanks

bridge/listener/rootchain_selfheal.go
```diff
@@ -69,6 +82,99 @@ func (rl *RootChainListener) startSelfHealing(ctx context.Context) {
...
+	}
+
+	// Get checkpoint parameters to get ChildChainBlockInterval.
+	checkpointParams, err := util.GetCheckpointParams(rl.cliCtx.Codec)
+	if err != nil {
+		rl.Logger.Error("Failed to get checkpoint params", "error", err)
+		return
+	}
+
+	l1HeaderBlockId = l1HeaderBlockId / checkpointParams.ChildChainBlockInterval
```
Raneet10, line 108: Can't remember but is this always supposed to be divisible by `ChildChainBlockInterval` ?

Raneet10 (approved):

bridge/listener/rootchain_selfheal.go
kamuikatsurgi: Yes, cause we multiply it by `ChildChainBlockInterval` in the smart contracts.

See [this](https://github.com/0xPolygon/pos-contracts/blob/10caf4342f54290e0d3cb6d763f6c8e3fa38374e/contracts/root/RootChainStorage.sol#L33-L36).

kamuikatsurgi: The name is different, between here and there, but yes that is always the case here.

kamuikatsurgi: We add it, not multiply, but yes, it will be the case, [see](https://github.com/0xPolygon/pos-contracts/blob/10caf4342f54290e0d3cb6d763f6c8e3fa38374e/contracts/root/RootChain.sol#L49)

PR #343, opened 2025-07-08
update docs with final v2 version

marcello33:
# Description

update docs with final v2 version


sophoah:
hey @marcello33 @avalkov  can we have an updated commit of the new sha1 of migrate.sh ?

marcello33:
Hi @sophoah 
It'd be useless to produce a new checksum of the the script right now.
Once we internally produce the genesis and its checksum, we'd need to update these vars in the script:
```
V1_GENESIS_CHECKSUM="..."
V2_GENESIS_CHECKSUM="..."
```
Only then the final checksum can be generated

PR #358, opened 2025-07-11
opt: GetEventRecordList and GetEventRecordCount

kamuikatsurgi:
# Description

@avalkov said this [api](https://heimdall-api.polygon.technology/clerk/event-records/list?page=100000) call loads the full db into memory, which is quite inefficient.

GetEventRecordList
- Before: Iterated through 5M records to reach page 100,000
- After: Uses collections.Range to jump directly to the needed records

GetEventRecordCount
- Before: Loaded all records into memory to count them
- After: Uses an iterator to count without loading record data


ssandeep (approved):

kamuikatsurgi:
Please review and approve the PR, but don't merge it. I want to test this on a mainnet sentry and validator node, just to be sure.

avalkov (approved):

marcello33 (approved):

PR #373, opened 2025-07-15
Add cache for checkpoint validation

avalkov:
# Description

Cache the check if block exists and the root hash.

### DIFF
--- a/common/cache/cache.go
+++ b/common/cache/cache.go
@@ -42,13 +42,13 @@ func (c *Cache[T]) Get(key string) (T, error) {
 	var zero T
 
 	if !found {
-		c.mu.Lock()
-		delete(c.items, key)
-		c.mu.Unlock()
-		return zero, errors.New("item expired")
+		return zero, errors.New("item not found")
 	}
 
 	if time.Now().UnixNano() > item.Expiration {
+		c.mu.Lock()
+		delete(c.items, key)
+		c.mu.Unlock()
 		return zero, errors.New("item expired")
 	}
 
--- a/common/cache/cache.go
+++ b/common/cache/cache.go
@@ -1,0 +1,53 @@
+package cache
+
+import (
+	"errors"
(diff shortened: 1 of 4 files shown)


x/checkpoint/keeper/side_msg_servers_test.go
```diff
@@ -76,15 +76,15 @@ func (s *KeeperTestSuite) TestSideHandleMsgCheckpoint() {
...
+			checkpoint.StartBlock+1,
 			checkpoint.EndBlock,
 			checkpoint.RootHash,
 			checkpoint.RootHash,
 			borChainId,
 		)
 
 		contractCaller.On("CheckIfBlocksExist", checkpoint.EndBlock+borChainTxConfirmations).Return(true, nil)
-		contractCaller.On("GetRootHash", checkpoint.StartBlock, checkpoint.EndBlock, uint64(1024)).Return(nil, nil)
+		contractCaller.On("GetRootHash", checkpoint.StartBlock+1, checkpoint.EndBlock, uint64(1024)).Return(nil, nil)
```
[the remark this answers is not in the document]
avalkov: To bypass the cache, these tests should fail but the cache has values from previous positive tests.

common/cache/cache.go
```diff
@@ -0,0 +1,53 @@
...
+	c.mu.RLock()
+	item, found := c.items[key]
+	c.mu.RUnlock()
+
+	var zero T
+	if !found || time.Now().UnixNano() > item.Expiration {
+		if found {
+			c.mu.Lock()
+			delete(c.items, key)
+			c.mu.Unlock()
```
kamuikatsurgi, line 47: Note: Here we can return the item expired error and outside of `if found`, we can return the item not found. This way we could understand more of what's happening.

avalkov: Done.

PR #393, opened 2025-07-24
fix(listener): fallback to correct ChainID when cliCtx.ChainID is unset

SunSet0418:
# Description

HeimdallListener was using an internally created `client.Context` that does not inherit the `ChainID` set via CLI (`initRootCmd`).  
As a result, `cliCtx.ChainID` remains an empty string (`""`) when `fetchFromAndToBlock()` is called, causing `GetInitialBlockHeight()` to return `0`.  
This results in unnecessary polling from block 0 and repeated 2-minute `CommitTimeout` errors in `GetBeginBlockEvents()`, ultimately preventing checkpoint submission.

This PR resolves the issue by using `nodeStatus.NodeInfo.Network` as a fallback value when `cliCtx.ChainID` is not set.

# Changes

# Breaking changes

N/A

# Node audience

This change is safe for all nodes. No flags or conditional logic added.

# Checklist

- [ ] I have added at least two reviewers or the whole pos-v1 team
- [x] I have added sufficient documentation in code (inline comments explaining fallback logic)

# Cross-repository changes

# Testing

- [ ] I have added unit tests

### Manual tests

- Stopped local Heimdall node and removed bridge DB
- Ensured no `--chain-id` was passed via CLI
- Started node and verified fallback to `nodeStatus.NodeInfo.Network` correctly sets `cliCtx.ChainID`
- Confirmed `fetchFromAndToBlock()` begins from expected block height, not `0`
- Confirmed checkpoints are now submitted correctly without any `CommitTimeout` logs

# Additional comments

The previous assumption that the `cliCtx` inside `HeimdallListener` would inherit values from the top-level CLI context does not hold true, as the listener initializes its own `client.Context`. This fix ensures checkpoint liveness even in scenarios where bridge DB is empty and CLI config is minimal.

### DIFF
--- a/bridge/listener/heimdall.go
+++ b/bridge/listener/heimdall.go
@@ -119,9 +119,6 @@ func (hl *HeimdallListener) StartPolling(ctx context.Context, pollInterval time.
 }
 
 func (hl *HeimdallListener) fetchFromAndToBlock(ctx context.Context) (uint64, uint64, error) {
-	fromBlock := uint64(0)
-	toBlock := uint64(0)
-
 	nodeStatus, err := helper.GetNodeStatus(hl.cliCtx, ctx)
 	if err != nil {
 		hl.Logger.Error("Error while fetching heimdall node status", "error", err)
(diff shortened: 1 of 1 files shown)


bridge/listener/heimdall.go
```diff
@@ -118,17 +119,27 @@ func (hl *HeimdallListener) StartPolling(ctx context.Context, pollInterval time.
...
+
 	nodeStatus, err := helper.GetNodeStatus(hl.cliCtx, ctx)
 	if err != nil {
 		hl.Logger.Error("Error while fetching heimdall node status", "error", err)
-		return fromBlock, toBlock, err
+		return 0, 0, err
+	}
+
+	chainId := hl.cliCtx.ChainID
+	if chainId == "" {
```
Raneet10, line 132: Hmm I wonder if we can set or inherit the chain ID value when initalizing the bridge. The fallback check is good to have though.

SunSet0418:
Hi @avalkov ,

I’m not entirely sure if your comments were suggesting this specific change,
but I’ve made an update that I believe reflects your intent.

Could you please check whether the following change aligns with your suggestion?

https://github.com/0xPolygon/heimdall-v2/pull/393/commits/b7ec3eab8eaf4a7dde3adf6408936911ffe0238

bridge/listener/heimdall.go
Raneet10: Right. I think we can fix this later. We'll track this internally, thanks!

avalkov (approved):

PR #394, opened 2025-07-24
Dockerfile: use multi-stage build optimisation

kamuikatsurgi:
# Description

Switched to Go 1.24‑alpine & Alpine 3.22; adopted a two‑stage build with SSH+cache mounts for go mod download and make build, then copied only the heimdalld binary and sh entrypoint into a minimal Alpine runtime (just certs, no Bash)—trimming the final image from hundreds of megabytes down to mere megabytes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,25 +1,41 @@
-FROM golang:latest
+# ─── BUILDER STAGE ───────────────────────────────────────────────────────────────
+FROM golang:1.24-alpine AS builder
 
-ARG HEIMDALL_DIR=/var/lib/heimdall
-ENV HEIMDALL_DIR=$HEIMDALL_DIR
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,25 +1,41 @@
...
 COPY . .
 
-RUN make build && cp build/heimdalld /usr/local/bin/
+RUN --mount=type=ssh \
+    --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    make build
+
+# ─── RUNTIME STAGE ────────────────────────────────────────────────────────────────
+FROM alpine:3.22
```
sreevalsanmr, line 25: change to `alpine:latest`

kamuikatsurgi: Will change that.

avalkov (approved):

PR #399, opened 2025-07-25
Fix import-validator-key command

Raneet10:
# Description

This PR fixes the `import-validator-key` command by properly registering the crypto types prior to importing. 

# Changes

# Checklist

- [x] I have added at least two reviewers or the whole pos-v1 team

## Testing

- [ ] I have added unit tests
- [x] I have tested this code manually on the local environment

### DIFF
--- a/cmd/heimdalld/cmd/commands.go
+++ b/cmd/heimdalld/cmd/commands.go
@@ -19,6 +19,7 @@ import (
 	cmtcmd "github.com/cometbft/cometbft/cmd/cometbft/commands"
 	cmtcfg "github.com/cometbft/cometbft/config"
 	"github.com/cometbft/cometbft/crypto"
+	cmcrypto "github.com/cometbft/cometbft/crypto"
 	"github.com/cometbft/cometbft/crypto/secp256k1"
 	cmtos "github.com/cometbft/cometbft/libs/os"
 	"github.com/cometbft/cometbft/p2p"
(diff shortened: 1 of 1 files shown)


cmd/heimdalld/cmd/commands.go
```diff
@@ -482,6 +482,10 @@ func generateValidatorKey() *cobra.Command {
 // importValidatorKey imports validator private key from the given file path
 func importValidatorKey() *cobra.Command {
 	cdc := codec.NewLegacyAmino()
+	cdc.RegisterInterface((*crypto.PubKey)(nil), nil)
+	cdc.RegisterInterface((*crypto.PrivKey)(nil), nil)
+	cdc.RegisterConcrete(secp256k1.PubKey{}, "tendermint/PubKeySecp256k1", nil)
```
avalkov, line 487: Should be `cometbft/PubKeySecp256k1eth`

cmd/heimdalld/cmd/commands.go
```diff
@@ -482,6 +482,10 @@ func generateValidatorKey() *cobra.Command {
...
+	cdc.RegisterConcrete(secp256k1.PrivKey{}, "tendermint/PrivKeySecp256k1", nil)
```
avalkov, line 488: Should be `cometbft/PrivKeySecp256k1eth`

cmd/heimdalld/cmd/commands.go
Raneet10: I believe we still use the `tendermint` namespace no ? I think the plan is to migrate to the new naming when we switch to upstream comet.

cmd/heimdalld/cmd/commands.go
Raneet10: Check the above comment.

cmd/heimdalld/cmd/commands.go
Raneet10: Okay seems I missed the detail where we asked everyone to also migrate the name. Updated [here](https://github.com/0xPolygon/heimdall-v2/pull/399/commits/a66b224eaa4ae07bf73789537434885829a1983b)

avalkov (approved):

PR #409, opened 2025-07-29
Re-enable voting power check in tally votes

Raneet10:
# Description

This PR re-enables the clause to verify the voting power coming from the vote extensions doesn't exceed the total power of the validator set (stored in stake module) in `tallyVotes`. Since CometBFT takes 2 heights for validator set changes to take effect, heimdall needs to keep track of the penultimate valset to tally correctly. 

# Changes

# Checklist

- [x] I have added at least two reviewers or the whole pos-v1 team

## Testing

- [ ] I have added unit tests

### DIFF
--- a/helper/config.go
+++ b/helper/config.go
@@ -444,7 +444,7 @@ func InitHeimdallConfigWith(homeDir string, heimdallConfigFileFromFlag string) {
 		veblopHeight = 0
 		tallyFixHeight = 0
 		disableVPCheckHeight = 0
-		disableValSetCheckHeight = 0 // TODO: confirm with team
+		disableValSetCheckHeight = 0
 		initialHeight = 0
 	}
 }


app/abci.go
```diff
@@ -115,6 +149,39 @@ func (app *HeimdallApp) NewProcessProposalHandler() sdk.ProcessProposalHandler {
...
+		var (
+			validatorSet *stakeTypes.ValidatorSet
+			err          error
+		)
+
+		chain := helper.GetConfig().Chain
+		initialHeight := int64(0)
+
+		// this is for UTs that have non zero initial height
+		if chain != helper.MainChain && chain != helper.AmoyChain && chain != helper.MumbaiChain {
```
avalkov, line 161: Same like above, also seems like a lot of code repetition, we can move it to separate function.

app/abci.go
```diff
@@ -590,10 +657,34 @@ func (app *HeimdallApp) PreBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlo
...
 
-	validatorSet, err := getPreviousBlockValidatorSet(ctx, app.StakeKeeper)
-	if err != nil {
-		logger.Error("Error occurred while getting previous block validator set", "error", err)
-		return nil, err
+	chain := helper.GetConfig().Chain
+	initialHeight := int64(0)
+
+	// this is for UTs that have non zero initial height
+	if chain != helper.MainChain && chain != helper.AmoyChain && chain != helper.MumbaiChain {
```
avalkov, line 664: Same like above.

app/vote_ext_utils.go
```diff
@@ -498,8 +489,10 @@ func checkNonRpVoteExtensionsSignatures(ctx sdk.Context, extVoteInfo []abciTypes
 
 		_, validator := validatorSet.GetByAddress(valAddr)
 		if validator == nil {
+			if ctx.BlockHeight() < helper.GetDisableValSetCheckHeight() || ctx.BlockHeight() >= helper.GetTallyFixHeight() {
```
avalkov, line 492: I think we can move this check in separate function and use it from all places and I think it should be `ctx.BlockHeight() >= helper.GetDisableValSetCheckHeight() && ctx.BlockHeight() < helper.GetTallyFixHeight()`

app/vote_ext_utils.go
```diff
@@ -194,12 +202,13 @@ func tallyVotes(extVoteInfo []abciTypes.ExtendedVoteInfo, logger log.Logger, tot
...
 		// calculate the total voting power in the voteMap
-		// power := voteMap[sidetxs.Vote_VOTE_YES] + voteMap[sidetxs.Vote_VOTE_NO] + voteMap[sidetxs.Vote_UNSPECIFIED]
+		power := voteMap[sidetxs.Vote_VOTE_YES] + voteMap[sidetxs.Vote_VOTE_NO] + voteMap[sidetxs.Vote_UNSPECIFIED]
+
 		// ensure the total votes do not exceed the total voting power
-		// if power > totalVotingPower {
-		// 	logger.Error("the votes power exceeds the total voting power", "txHash", txHash, "power", power, "totalVotingPower", totalVotingPower)
-		// 	return nil, nil, nil, fmt.Errorf("votes power %d exceeds total voting power %d for txHash %s", power, totalVotingPower, txHash)
-		// }
+		if (currentHeight < helper.GetDisableVPCheckHeight() && currentHeight >= helper.GetTallyFixHeight()) && power > totalVotingPower {
```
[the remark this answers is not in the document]
Raneet10: Right noticed that during testing.

app/abci.go
```diff
@@ -33,12 +34,45 @@ func (app *HeimdallApp) NewPrepareProposalHandler() sdk.PrepareProposalHandler {
...
+		var (
+			validatorSet *stakeTypes.ValidatorSet
+			err          error
+		)
+
+		chain := helper.GetConfig().Chain
+		initialHeight := int64(0)
+
+		// this is for UTs that have non zero initial height
+		if chain != helper.MainChain && chain != helper.AmoyChain && chain != helper.MumbaiChain {
```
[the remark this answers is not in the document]
Raneet10: Okay but those chains are running so they don't have it stored so not gonna matter I think. Agreed on the second point.

app/vote_ext_utils.go
avalkov: Yea, your right.

x/milestone/abci/abci.go
```diff
@@ -476,3 +479,7 @@ func ValidateMilestoneProposition(ctx sdk.Context, milestoneKeeper *keeper.Keepe
 
 	return nil
 }
+
+func ShouldErrorOnValidatorNotFound(ctx sdk.Context) bool {
+	return ctx.BlockHeight() >= helper.GetTallyFixHeight() || ctx.BlockHeight() < helper.GetDisableValSetCheckHeight()
```
jmalicevic, line 484: Is there a relationship between `GetTallyFixHeight` and `GetDisableValSetCheckHegith`? Should the former not be greater than the second and in fact, should it not be `GetDisableValSetCheckHegith` + 1?

Raneet10: Yeah so `GetDisableValSetCheckHegith` would be the height we disabled the check and `GetTallyFixHeight` would be the height where it'd be re-enabed.

jmalicevic: Ah I see. And there should be an error on validator not found only before the height at which you had disabled this and once the proper validator counting starts happening. Makes sense.

x/stake/keeper/abci.go
```diff
@@ -4,15 +4,16 @@ import (
...
-	"github.com/0xPolygon/heimdall-v2/x/stake/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
 // EndBlocker called at the end of every block and returns the validator updates
 func (k *Keeper) EndBlocker(ctx context.Context) ([]abci.ValidatorUpdate, error) {
 	k.PanicIfSetupIsIncomplete()
 	defer telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), telemetry.MetricKeyEndBlocker)
-	return k.ApplyAndReturnValidatorSetUpdates(ctx)
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
```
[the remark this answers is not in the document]
sergio-mena: I think it would be safer to unwrap the context inside method `ApplyAndReturnValidatorSetUpdates`.
The reason is that `ApplyAndReturnValidatorSetUpdates` first calls `GetValidatorSet` (line 527): this code hasn't changed and was passing the ctx without unwrapping. I would keep it that way, and only unwrap the context after that call to `GetValidatorSet` in line 527.
Note that I didn't look into `GetValidatorSet` (if that method is able to receive a context both with and without unwrapping, then my comment is superfluous).

x/chainmanager/keeper/genesis.go
```diff
@@ -5,13 +5,19 @@ import (
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
 // InitGenesis sets chainmanager information for genesis.
 func (k Keeper) InitGenesis(ctx context.Context, data *types.GenesisState) {
 	if err := k.SetParams(ctx, data.Params); err != nil {
 		panic(fmt.Errorf("failed to set chainmanager params: %w", err))
 	}
+
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
```
sergio-mena, line 17: Hmmm.... if the migration is already done, are we sure this code will be hit?

The only way I see this code could ever be hit, is if a node was started from scratch, and was made to block-synch all the way to the tip (still feasible as the chain is only a few weeks old).

HOWEVER: Bear in mind that you're setting a value in the avl DB. So if a node starting from scratch does this, but the actual nodes that participated in the chain migration last month didn't do it, the node will halt with an app mismatch on the first block it tries to apply.
Have you tested this in a devnet/testnet?

app/vote_ext_utils.go
```diff
@@ -700,3 +702,7 @@ func packExtensionWithVote(extension []byte) []byte {
 	buf.Write(extension)
 	return buf.Bytes()
 }
+
+func shouldCheckVotingPower(currentHeight int64) bool {
+	return currentHeight < helper.GetDisableVPCheckHeight() || currentHeight >= helper.GetTallyFixHeight()
```
sergio-mena, line 707: Should it be `helper.GetTallyFixHeight() + 1`?  (not sure, just asking)

x/stake/keeper/validator.go
```diff
@@ -494,15 +522,31 @@ func (k *Keeper) ValidatorAddressCodec() addresscodec.Codec {
...
+func (k Keeper) ApplyAndReturnValidatorSetUpdates(ctx sdk.Context) (updates []abci.ValidatorUpdate, err error) {
 	var cmtValUpdates []abci.ValidatorUpdate
 	currentValidatorSet, err := k.GetValidatorSet(ctx)
 	if err != nil {
 		k.Logger(ctx).Error("error while calling the GetValidatorSet fn", "err", err)
 		return nil, err
 	}
 
-	// save the previous block's validator set
+	if ctx.BlockHeight() >= helper.GetTallyFixHeight()-1 {
```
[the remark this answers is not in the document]
sergio-mena: I think I now get why you need the "id". It's because we want to allow different validators to update at different heights (rather than an SDK-like coordinated update, where everyone upgrades at the same height).
Did I get it right?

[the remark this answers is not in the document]
Raneet10: Well it's because we want to ensure every node initializes the penultimate validator set one block prior to hitting the `GetTallyFixHeight()`, from which we activate the new logic of using the second last validator set. And we must make it consistent across different validators, hence the if statement. Let me know if this makes it clear.

app/vote_ext_utils.go
Raneet10: Well `GetTallyFixHeight()` is the height at which we re-activate the checks regarding voting power and validator set. Why would you think it should be `GetTallyFixHeight() + 1` ?

app/abci.go
```diff
@@ -33,12 +34,18 @@ func (app *HeimdallApp) NewPrepareProposalHandler() sdk.PrepareProposalHandler {
 	return func(ctx sdk.Context, req *abci.RequestPrepareProposal) (*abci.ResponsePrepareProposal, error) {
 		logger := app.Logger()
 
-		if err := ValidateVoteExtensions(ctx, req.Height, req.LocalLastCommit.Votes, req.LocalLastCommit.Round, app.StakeKeeper, app.MilestoneKeeper); err != nil {
+		validatorSet, err := app.getValidatorSetForHeight(ctx, req.Height)
+		if err != nil {
+			logger.Error("Error occurred while getting validator set for height in PrepareProposal", "error", err)
```
marcello33, line 39: Shall we print the height too?

app/abci.go
```diff
@@ -115,6 +122,12 @@ func (app *HeimdallApp) NewProcessProposalHandler() sdk.ProcessProposalHandler {
 	return func(ctx sdk.Context, req *abci.RequestProcessProposal) (*abci.ResponseProcessProposal, error) {
 		logger := app.Logger()
 
+		validatorSet, err := app.getValidatorSetForHeight(ctx, req.Height)
+		if err != nil {
+			logger.Error("Error occurred while getting validator set for height in ProcessProposal", "error", err)
```
marcello33, line 127: And here as well? Just to help in case of debugging

app/abci.go
```diff
@@ -590,9 +603,9 @@ func (app *HeimdallApp) PreBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlo
 		return nil, err
 	}
 
-	validatorSet, err := getPreviousBlockValidatorSet(ctx, app.StakeKeeper)
+	validatorSet, err := app.getValidatorSetForHeight(ctx, req.Height)
 	if err != nil {
-		logger.Error("Error occurred while getting previous block validator set", "error", err)
+		logger.Error("Error occurred while getting validator set for height in PreBlocker", "error", err)
```
marcello33, line 608: And here as well

app/abci.go
```diff
@@ -812,3 +826,45 @@ func (app *HeimdallApp) PreBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlo
...
+func (app *HeimdallApp) getInitialHeight(ctx sdk.Context) (int64, error) {
+	chain := helper.GetConfig().Chain
+	initialHeight := int64(0)
+	var err error
+
+	// this is for UTs that have non zero initial height
+	if chain != helper.MainChain && chain != helper.AmoyChain && chain != helper.MumbaiChain {
+		initialHeight, err = app.ChainManagerKeeper.GetInitialChainHeight(ctx)
+		if err != nil {
+			return -1, fmt.Errorf("error occurred while getting initial chain height: %w", err)
```
marcello33, line 839: So this is called by `getValidatorSetForHeight` and calling points are the ABCI methods. Wondering if it's better to return 0 and no error in case (to avoid panics).

x/stake/keeper/validator.go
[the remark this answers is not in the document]
sergio-mena: Yeah, matches my understanding. Just to confirm: you will be allowing the different nodes to upgrades at different heights, within a pre-agreed time span (e.g. a week). I think I remember @avalkov explaining that in the last meeting I took part in.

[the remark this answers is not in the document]
avalkov: @sergio-mena We will trigger the upgrade at same pre-agreed height for all nodes, we will calculate roughly a height that is 1-2 weeks in the future.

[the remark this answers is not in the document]
sergio-mena: Thanks @avalkov , when you say "we will trigger" you mean that the new logic will kick in at that predetermined height. But you will nevertheless allow validators to upgrade the binary a few days earlier. Is my understanding right?

x/chainmanager/types/keys.go
```diff
@@ -13,4 +13,5 @@ const (
 	RouterKey = ModuleName
 )
 
-var ParamsKey = collections.NewPrefix(0) // ParamsKey is the key to store the params in the store
+var ParamsKey = collections.NewPrefix(0)             // ParamsKey is the key to store the params in the store
+var InitialChainHeightKey = collections.NewPrefix(1) // InitialChainHeightKey is the key to store the initial chain height in the store
```
sergio-mena, line 17: I looked into how/where this is set and/or used and and didn't find any instance of those (I mean, no `Get` or `Set`).
Since I see you're now providing the initial height via the helper, is this variable, and the corresponding field in the Keep necessary? If not, I'd remove it

app/abci.go
```diff
@@ -812,3 +826,25 @@ func (app *HeimdallApp) PreBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlo
...
 	return app.ModuleManager.PreBlock(ctx)
 }
+
+func (app *HeimdallApp) getValidatorSetForHeight(ctx sdk.Context, height int64) (*stakeTypes.ValidatorSet, error) {
+	var (
+		validatorSet *stakeTypes.ValidatorSet
+		err          error
+	)
+
+	if height >= helper.GetTallyFixHeight() && height >= helper.GetInitialHeight()+2 {
```
sergio-mena, line 836: Why do we need `height >= helper.GetInitialHeight()+2` in this IF?

IIUC, `TallyFixHeight` is way greater than `InitialHeight`... so the second condition is trivially true if the first one is true

x/milestone/abci/abci.go
```diff
@@ -476,3 +479,7 @@ func ValidateMilestoneProposition(ctx sdk.Context, milestoneKeeper *keeper.Keepe
 
 	return nil
 }
+
+func ShouldErrorOnValidatorNotFound(ctx sdk.Context) bool {
```
sergio-mena, line 483: Very minor: as you only use `ctx.BlockHeight()`, consider passing the height directly to the function as a parameter, rather than the whole context.

app/vote_ext_utils.go
```diff
@@ -547,8 +534,10 @@ func getMajorityNonRpVoteExtension(ctx sdk.Context, extVoteInfo []abciTypes.Exte
 
 		_, validator := validatorSet.GetByAddress(valAddr)
 		if validator == nil {
+			if milestoneAbci.ShouldErrorOnValidatorNotFound(ctx) {
```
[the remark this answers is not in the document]
sergio-mena: WAIT A MINUTE -- I think there's a bug below in this function (which predates this PR).
The `for` loop in lines 552-557 below is iterating over a map. Iterating over golang maps is non-deterministic (it may change depending on CPU architecture or go version) so this might bite you back badly in the future.
So, what is the bug?  Let's say that -- for whatever reason -- there are two entries in the map with equal voting power, and that that voting power value is the maximum ... boom: non-determinism!

Please do the following instead of iterating directly over the map:
* Extract the map's keys -- TX hashes -- into a slice
* Sort that slice lexicographically
* Iterate over the sorted slice, rather than the map directly

x/chainmanager/types/keys.go
Raneet10: Yeah right I missed removing the leftover references.

app/abci.go
Raneet10: Yes this is true for production but for unit tests and devnets we need to have this condition.

app/vote_ext_utils.go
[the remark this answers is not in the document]
Raneet10: True but for a node syncing from scratch, this check was present prior to us disabling it at `GetDisableVPCheckHeight()`. Although I suppose this condition never failed before the incident, which effectively would function as if the condition were disabled, I'd still love to keep this for determinism and logical correctness. LMK what you think.

x/milestone/abci/abci.go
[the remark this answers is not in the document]
Raneet10: Same comment as [above](https://github.com/0xPolygon/heimdall-v2/pull/409#discussion_r2259028776)

app/vote_ext_utils.go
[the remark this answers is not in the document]
Raneet10: Hmm yeah indeed an extremely unlikely condition. We need not consider this right ? It would just further complicate the flow and might add extra overhead.

x/milestone/abci/abci.go
```diff
@@ -177,8 +178,10 @@ func GetMajorityMilestoneProposition(
 
 		_, validator := validatorSet.GetByAddress(valAddr)
 		if validator == nil {
+			if ShouldErrorOnValidatorNotFound(ctx) {
```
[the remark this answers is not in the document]
Raneet10: Same comment as [above](https://github.com/0xPolygon/heimdall-v2/pull/409#discussion_r2259091781).

app/vote_ext_utils.go
```diff
@@ -560,8 +549,15 @@ func getMajorityNonRpVoteExtension(ctx sdk.Context, extVoteInfo []abciTypes.Exte
...
 	var maxVotingPower int64
 	var maxHash string
-	for hash, votingPower := range hashToVotingPower {
-		if votingPower > maxVotingPower {
+
+	hashList := make([]string, 0, len(hashToVotingPower))
+	for hash := range hashToVotingPower {
+		hashList = append(hashList, hash)
+	}
+	sort.Strings(hashList)
```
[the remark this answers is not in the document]
Raneet10: It is a good practice when iterating over maps deterministically.

x/stake/keeper/validator.go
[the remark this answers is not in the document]
Raneet10: Yes that's correct. We ask all validators/operators to update their binaries beforehand so that they don't crash when the new logic kicks in from the "hardfork" height.

helper/config.go
```diff
@@ -1151,3 +1161,8 @@ func SetTestPrivPubKey(privKey secp256k1.PrivKey) {
 	}
 	pubKeyObject = pubKey
 }
+
+// SetTestInitialHeight sets test the initial height for testing
+func SetTestInitialHeight(height int64) {
+	initialHeight = height
+}
```
sergio-mena, line 1168: <picky-mode>
I would be good to move this function to file `helper/config_test.go`, to avoid the risk of calling it in production by mistake.
</picky-mode>

sergio-mena (approved):
🚀🚀🚀

helper/config.go
```diff
@@ -416,12 +424,28 @@ func InitHeimdallConfigWith(homeDir string, heimdallConfigFileFromFlag string) {
...
 		veblopHeight = 0
+		tallyFixHeight = 0           // TODO: TBD
+		disableVPCheckHeight = 0     // TODO: confirm with team
+		disableValSetCheckHeight = 0 // TODO: confirm with team
+		initialHeight = 0
 	default:
 		veblopHeight = 0
+		tallyFixHeight = 0
+		disableVPCheckHeight = 0
+		disableValSetCheckHeight = 0 // TODO: confirm with team
```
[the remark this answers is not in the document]
Raneet10: Done

PR #425, opened 2025-08-14
update: cosmos-sdk

kamuikatsurgi:
# Description

Bumps cosmos-sdk versions.


go.mod
```diff
@@ -325,16 +324,16 @@ replace (
 	cosmossdk.io/api => github.com/0xPolygon/cosmos-sdk/api v0.7.6-0.20250429154832-7177ebac408e
 	cosmossdk.io/client/v2 => github.com/0xPolygon/cosmos-sdk/client/v2 v2.0.0-beta.6
 	cosmossdk.io/collections => github.com/0xPolygon/cosmos-sdk/collections v0.4.0
-	cosmossdk.io/core => github.com/0xPolygon/cosmos-sdk/core v0.11.3-0.20241126102051-89dc71d02611
-	cosmossdk.io/errors => github.com/0xPolygon/cosmos-sdk/errors v1.0.0-beta.7.0.20241126102051-89dc71d02611
+	cosmossdk.io/core => github.com/0xPolygon/cosmos-sdk/core v0.11.3-0.20250814084204-6e6b8f232b2a
```
kamuikatsurgi, line 327: I have updated the commit hashes here to the latest commit on the `devel` branch of the `cosmos-sdk`. Let me know if I wasn't supposed to do this.

pratikspatil024 (approved):

go.mod
kamuikatsurgi: We'll have to upgrade sometime anyways, better to keep up with it regularly.

kamuikatsurgi: Hmmm...yeah, that's a good point. Maybe we can deploy this somewhere and check if it doesn't break something? I'll deploy this branch to an amoy validator node and let it run for sometime. Thanks for the explanation though.

PR #437, opened 2025-08-22
fix: update lastSeqNo if less than account sequence

kamuikatsurgi:
# Description

Addresses a race condition where the bridge was not using the updated sequence number when doing tx from the CLI.

### DIFF
--- a/bridge/broadcaster/broadcaster.go
+++ b/bridge/broadcaster/broadcaster.go
@@ -71,8 +71,8 @@ func NewTxBroadcaster(
 			case <-ctx.Done():
 				// return a minimal broadcaster so caller can shut down cleanly
 				return &TxBroadcaster{
-					logger: logger,
 					CliCtx: cliCtx.WithFromAddress(fromAddr),
+					logger: logger,
 				}
 			default:
 			}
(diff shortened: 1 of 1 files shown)


app/vote_ext_utils.go
```diff
@@ -201,29 +201,23 @@ func tallyVotes(extVoteInfo []abciTypes.ExtendedVoteInfo, logger log.Logger, tot
...
-			// skipped - 2/3 consensus not reached.
 			logger.Debug("Skipped side-tx", "txHash", txHash)
 			consensus.RecordConsensusFailure()
-			// append to rejected tx slice
 			skippedTxs = append(skippedTxs, common.FromHex(txHash))
 		}
 	}
 
-	logger.Debug(fmt.Sprintf("Height %d: approved %d txs, rejected %d txs, skipped %d txs. ", currentHeight, len(approvedTxs), len(rejectedTxs), len(skippedTxs)))
+	logger.Info("Tallied votes", "height", currentHeight, "approvedTxs", len(approvedTxs), "rejectedTxs", len(rejectedTxs), "skippedTxs", len(skippedTxs))
```
Raneet10, line 218: Why bump the log level ?

marcello33 (approved):

PR #440, opened 2025-08-23
Stateless sync

cffls:
# Description

Merge stateless_sync to develop.

# Changes

# Breaking changes

Please complete this section if any breaking changes have been made, otherwise delete it

# Node audience

In case this PR includes changes that must be applied only to a subset of nodes, please specify how you handled it
(e.g., by adding a flag with a default value...)

# Checklist

- [ ] I have added at least two reviewers or the whole pos-v1 team
- [ ] I will be resolving comments — if any — by pushing each fix in a separate commit and linking the commit hash in the comment reply

# Cross-repository changes

  - In case link the PR here:
  - In case link the PR here:

## Testing

- [ ] I have added unit tests

### DIFF
--- a/helper/config.go
+++ b/helper/config.go
@@ -446,7 +446,7 @@ func InitHeimdallConfigWith(homeDir string, heimdallConfigFileFromFlag string) {
 		disableValSetCheckHeight = 10618299
 		initialHeight = 8788501
 	default:
-		veblopHeight = 384 // VeBlop height for local testnets.
+		veblopHeight = 256 // VeBlop height for local testnets.
 		tallyFixHeight = 0
 		disableVPCheckHeight = 0
 		disableValSetCheckHeight = 0


app/abci.go
```diff
@@ -738,9 +742,30 @@ func (app *HeimdallApp) PreBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlo
 		}
 		msCache.Write()
 	} else {
-		if err := app.checkAndRotateCurrentSpan(ctx); err != nil {
+		// If we can't reach the 2/3 majority, we need to check if the there is at least 1/3 of the voting power supporting a new milestone
+		minMajorityVP := totalVotingPower/3 + 1
```
lucca30, line 746: "At least 1/3" means `ceil(t/3)` which can be converted for the integer operation `(t+2)/3` so it covers a slightly bit more than `(t/3)+1`

app/abci.go
```diff
@@ -644,10 +644,14 @@ func (app *HeimdallApp) PreBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlo
 		lastEndHash = lastMilestone.Hash
 	}
 
+	totalVotingPower := validatorSet.GetTotalVotingPower()
+	majorityVP := totalVotingPower*2/3 + 1
```
lucca30, line 648: If we need a strict condition of `>2/3` we can think of `floor(2t/3) +1`. Since `floor(2t/3) = t - ceil(t/3)`, so we have:

```
t - ceil(t/3) + 1
t - ((t+2)/3) + 1

```

app/abci.go
```diff
@@ -644,10 +644,14 @@ func (app *HeimdallApp) PreBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlo
 		lastEndHash = lastMilestone.Hash
 	}
 
+	totalVotingPower := validatorSet.GetTotalVotingPower()
```
lucca30, line 647: is there any risk of 0 voting power?

app/abci.go
```diff
@@ -738,9 +742,30 @@ func (app *HeimdallApp) PreBlocker(ctx sdk.Context, req *abci.RequestFinalizeBlo
...
+			lastEndBlock,
+			lastEndHash,
+		)
+		if err != nil {
+			logger.Error("Error occurred while getting 33% majority milestone proposition", "error", err)
 			return nil, err
 		}
+
+		if pendingMilestone == nil {
+			if err := app.checkAndRotateCurrentSpan(ctx); err != nil {
```
lucca30, line 763: Shouldnt we use `addMilestoneCtx`? I saw some snippets in this function (event the ones you didn't touched) that may be interesting use because of consistency of the cacheWrite vs read. wdyt?

lucca30 (commented):
lgtm, just small changes to observe

app/abci.go
cffls: I doubt it could be 0. It means everyone unstaked and we can't even make new blocks in heimdall.

app/abci.go
cffls: As discussed, it will only make a difference when `T` is extremely small, e.g. `T=3`. In normal cases, `(t/3)+1` check is too strict but it won't hurt in reality. We will keep using this for readability.

app/abci.go
cffls: Same as above, we will keep the current approach for better readability.

lucca30 (approved):

PR #467, opened 2025-09-15
Default Prune and Compaction Config

lucca30:
# Description

Including new default prune and compaction flags to be inserted on Heimdall. These new configurations comes from the new introduced CometBFT features: https://github.com/0xPolygon/cometbft/pull/23 and https://github.com/0xPolygon/cometbft/pull/24

# Changes

### DIFF
--- a/cmd/heimdalld/cmd/sanitize.go
+++ b/cmd/heimdalld/cmd/sanitize.go
@@ -3,9 +3,10 @@ package heimdalld
 import (
 	"fmt"
 
-	"github.com/0xPolygon/heimdall-v2/helper"
 	"github.com/mitchellh/mapstructure"
 	"github.com/spf13/viper"
+
+	"github.com/0xPolygon/heimdall-v2/helper"
 )
 
 func SanitizeConfig(rootViper *viper.Viper) error {
--- a/cmd/heimdalld/cmd/sanitize.go
+++ b/cmd/heimdalld/cmd/sanitize.go
@@ -1,0 +1,31 @@
+package heimdalld
+
+import (
+	"fmt"
+
+	"github.com/0xPolygon/heimdall-v2/helper"
+	"github.com/mitchellh/mapstructure"
+	"github.com/spf13/viper"
+)
+
+func SanitizeConfig(rootViper *viper.Viper) error {
+	var appCfg helper.CustomAppConfig
+	decodeHook := mapstructure.ComposeDecodeHookFunc(
+		mapstructure.StringToTimeDurationHookFunc(),
+		mapstructure.StringToSliceHookFunc(","),
+	)
+	if err := rootViper.Unmarshal(&appCfg, viper.DecodeHook(decodeHook)); err != nil {
+		return fmt.Errorf("invalid config: %w", err)
+	}
+	if notes, kv := appCfg.Sanitize(); len(notes) > 0 {
+		fmt.Println("## Detected some configuration values to be sanitized")
+		for _, n := range notes {
+			fmt.Println("[config] adjusted:", n)
+		}
+		// write sanitized values back into Viper
+		for k, v := range kv {
+			rootViper.Set(k, v)
+		}
+	}
+	return nil
+}
--- a/helper/config.go
+++ b/helper/config.go
@@ -1156,8 +1156,8 @@ func GetBorGRPCClient() *borgrpc.BorGRPCClient {
 func (c *CustomAppConfig) Sanitize() (notes []string, kv map[string]any) {
 	kv = make(map[string]any)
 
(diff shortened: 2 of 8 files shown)


packaging/templates/config/amoy/config.toml
```diff
@@ -458,6 +458,28 @@ peer_query_maj23_sleep_duration = "200ms"
...
+compaction_interval = 1000
+
+[storage.pruning]
+
+# The time period between automated background pruning operations.
+interval = "10m0s"
+
+# Indexer pruning enabling. Note that this has not been tested in production
+# It should be enabled on full nodes to confirm it is not impacting performance. 
+indexer_pruning_enabled = false
```
marcello33, line 481: Should we mention this hasn't been tested? I guess it comes from comet.
Since we are testing this on amoy, the misleading comment should be removed imho

packaging/templates/config/mainnet/config.toml
```diff
@@ -458,6 +458,28 @@ peer_query_maj23_sleep_duration = "200ms"
...
+compaction_interval = 1000
+
+[storage.pruning]
+
+# The time period between automated background pruning operations.
+interval = "10m0s"
+
+# Indexer pruning enabling. Note that this has not been tested in production
+# It should be enabled on full nodes to confirm it is not impacting performance. 
+indexer_pruning_enabled = false
```
marcello33, line 481: Same comment as amoy config

packaging/templates/config/mainnet/app.toml
```diff
@@ -48,9 +48,11 @@ halt-time = 0
...
 #
 # Note: CometBFT block pruning is dependant on this parameter in conjunction
-# with the unbonding (safety threshold) period, state pruning and state sync
-# snapshot parameters to determine the correct minimum value of
+# with the unbonding (safety threshold) period, state pruning, state sync
+# snapshot parameters and evidence max age to determine the correct minimum value of
 # ResponseCommit.RetainHeight.
+# Since evidence.max_age_num_blocks is currently set on 100000, thats the minimum value
+# expected to be set
 min-retain-blocks = 0
```
marcello33, line 56: Just wondering what is the value we are proposing for our nodes

lucca30: `100000` sounds like a good choice imo.

marcello33: I think we should delegate the decision to node operators, and avoid an enforced value via HF
Ofc, we will provide suggested default config values.
When going on amoy/mainnet, I'd always keep 1-2 months of data.
Do you know what'd the size for that blocks window?

lucca30: In this case 2.5Mi seems the best choose to recommend assuming 1Block/s which means 2.5Mi is a month. Anyway, if someone tries to do less than that they would just reach the 100k value imposed currently by genesis.

lucca30: As mentioned by @avalkov , we can do this update via gov proposal:
https://github.com/0xPolygon/cosmos-sdk/blob/devel/x/gov/keeper/msg_validator.go#L35

lucca30: After team discussion we decided a config sanitization on `min-retain-blocks` would be the better approach. Implemented on last commit

cmd/heimdalld/cmd/sanitize.go
```diff
@@ -0,0 +1,32 @@
...
+		mapstructure.StringToTimeDurationHookFunc(),
+		mapstructure.StringToSliceHookFunc(","),
+	)
+	if err := rootViper.Unmarshal(&appCfg, viper.DecodeHook(decodeHook)); err != nil {
+		return fmt.Errorf("invalid config: %w", err)
+	}
+	if notes, kv := appCfg.Sanitize(); len(notes) > 0 {
+		fmt.Println("## Detected some configuration values to be sanitized")
+		for _, n := range notes {
+			fmt.Println("[config] adjusted:", n)
```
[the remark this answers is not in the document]
kamuikatsurgi: Move this:
```go
err = SanitizeConfig(serverCtx.Viper)
  if err != nil {
    return err
  }
```
after `logger, err := server.CreateSDKLogger(serverCtx, cmd.OutOrStdout())`.

helper/config.go
```diff
@@ -124,6 +124,8 @@ const (
 
 	// MaxStateSyncSize is the new max state sync size after SpanOverrideHeight hard fork
 	MaxStateSyncSize = 30000
+
+	MinimunMinRetainBlocks = 2500000
```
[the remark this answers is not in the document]
lucca30: Yeah, I was thinking about that. What about `MinimunEnforcedMinRetainBlocks`? Because default is 0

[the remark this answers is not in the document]
lucca30: deal

PR #472, opened 2025-09-24
Set producer planned downtime

avalkov:
# Description

Implemented tx to set block producer planned downtime.


x/bor/keeper/side_msg_server.go
```diff
@@ -344,6 +391,189 @@ func (s sideMsgServer) PostHandleMsgBackfillSpans(ctx sdk.Context, msgI sdk.Msg,
...
+			if cur.SelectedProducers[0].ValId == validatorId {
+				// At minimum, confirm the planned downtime applies to this producer over this span.
+				isDown, derr := s.k.IsProducerDownForBlockRange(ctx, cur.StartBlock, cur.EndBlock, validatorId)
+				if derr != nil {
+					logger.Error("failed to check producer downtime overlap", "spanId", cur.Id, "error", derr)
+					return derr
+				}
+
+				if isDown {
+					err = s.k.AddNewVeblopSpan(ctx, validatorId, lastSpan.EndBlock+1, lastSpan.EndBlock+params.SpanDuration, lastSpan.BorChainId, latestActiveProducer, uint64(ctx.BlockHeight()))
```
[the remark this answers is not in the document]
avalkov: Thanks for pointing out. I think this handler was overly complex, I completely reworked and simplified it.

x/bor/keeper/msg_server.go
```diff
@@ -259,6 +259,39 @@ func (s msgServer) BackfillSpans(ctx context.Context, msg *types.MsgBackfillSpan
 	return &types.MsgBackfillSpansResponse{}, nil
 }
 
+func (s msgServer) SetProducerDowntime(ctx context.Context, msg *types.MsgSetProducerDowntime) (*types.MsgSetProducerDowntimeResponse, error) {
+	var err error
+	start := time.Now()
+	defer recordBorTransactionMetric(api.ProducerDowntimeMethod, start, &err)
+
+	validators := s.sk.GetSpanEligibleValidators(ctx)
```
cffls: If the proposer is not part of eligible producer, we should reject immediately. This will avoid unnecessary computation in case a validator who is outside of producer set spamming the network with such proposals.

x/bor/keeper/side_msg_server.go
```diff
@@ -344,6 +391,181 @@ func (s sideMsgServer) PostHandleMsgBackfillSpans(ctx sdk.Context, msgI sdk.Msg,
...
+	if err != nil {
+		return err
+	}
+
+	if !foundOverlappingSpan {
+		// No spans overlap the downtime window; nothing to do.
+		return nil
+	}
+
+	if err := s.k.AddNewVeblopSpan(ctx, validatorId, msg.DowntimeRange.StartBlock, msg.DowntimeRange.StartBlock+params.SpanDuration, lastSpan.BorChainId, latestActiveProducer, uint64(ctx.BlockHeight())); err != nil {
```
cffls: Would it be better to set the end block of the new span as the block right before the next span planned previously? It would have less impact on the following span.

x/bor/keeper/veblop.go
```diff
@@ -237,7 +237,7 @@ func (k *Keeper) ClearLatestFailedProducer(ctx context.Context) error {
 
 // SelectNextSpanProducer selects the next producer for a new span.
 // It calculates candidate set, filters by active producers and selects one.
-func (k *Keeper) SelectNextSpanProducer(ctx context.Context, currentProducer uint64, activeValidatorIDs map[uint64]struct{}, producerSetLimit uint64) (uint64, error) {
+func (k *Keeper) SelectNextSpanProducer(ctx sdk.Context, currentProducer uint64, activeValidatorIDs map[uint64]struct{}, producerSetLimit, startBlock, endBlock uint64) (uint64, error) {
```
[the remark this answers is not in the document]
avalkov: I think in Rio there is no more span proposal tx and the generation of new spans happens directly in PreBlocker, right before execution of the transactions posthandlers.

app/app.go
```diff
@@ -506,6 +505,16 @@ func (app *HeimdallApp) CheckTx(req *abci.RequestCheckTx) (*abci.ResponseCheckTx
 						Log:  err.Error(),
 					}, nil
 				}
+			} else if _, ok := msg.(*borTypes.MsgSetProducerDowntime); ok {
+				// Create a context for validation
+				ctx := app.NewUncachedContext(true, cmtproto.Header{})
```
lucca30: Shouldn't we set here the next block? In this way its setting `Height : 0` so will always reject until we set downtime on genesis.

```
ctx := app.NewUncachedContext(true, cmtproto.Header{Height: app.LastBlockHeight() + 1})
```

x/bor/client/cli/tx.go
```diff
@@ -122,3 +125,132 @@ func NewSpanProposalCmd() *cobra.Command {
...
+			if err != nil {
+				return fmt.Errorf("producer address is invalid: %w", err)
+			}
+
+			startTimeUTC := viper.GetInt(FlagStartTimestampUTC)
+			if startTimeUTC <= 0 {
+				return fmt.Errorf("start timestamp utc is invalid")
+			}
+
+			endTimeUTC := viper.GetInt(FlagEndTimestampUTC)
```
lucca30: Trying to get a int but setting flag as string:

```
	cmd.Flags().String(FlagStartTimestampUTC, "", "--start-timestamp-utc=<start-timestamp-utc>")
	cmd.Flags().String(FlagEndTimestampUTC, "", "--end-timestamp-utc=<end-timestamp-utc>")

```

x/bor/client/cli/tx.go
```diff
@@ -122,3 +125,132 @@ func NewSpanProposalCmd() *cobra.Command {
...
+	}
+
+	if err := cmd.MarkFlagRequired(FlagEndTimestampUTC); err != nil {
+		fmt.Printf("NewProducerDowntimeCmd | MarkFlagRequired | FlagEndTimestampUTC Error: %v", err)
+	}
+
+	if err := cmd.MarkFlagRequired(flags.FlagChainID); err != nil {
+		fmt.Printf("NewProducerDowntimeCmd | MarkFlagRequired | FlagChainID Error: %v", err)
+	}
+
```
lucca30: Why forcing chainId? The node itself can already fill it.

Is it a strategy to force user double check what they are doing?

x/bor/client/cli/tx.go
```diff
@@ -122,3 +125,132 @@ func NewSpanProposalCmd() *cobra.Command {
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx, err := client.GetClientQueryContext(cmd)
+			if err != nil {
+				return err
+			}
+
+			producerAddress := viper.GetString(FlagProducerAddress)
+			if producerAddress == "" {
+				producerAddress = clientCtx.GetFromAddress().String()
+			}
```
lucca30: We have this default logic but we do also mark flag as required. So I think we should remove this fill logic

x/bor/client/cli/tx.go
```diff
@@ -122,3 +125,132 @@ func NewSpanProposalCmd() *cobra.Command {
...
+
+			borClient := helper.GetBorClient()
+
+			block, err := borClient.BlockByNumber(clientCtx.CmdContext, nil)
+			if err != nil {
+				return fmt.Errorf("failed to get latest bor block: %w", err)
+			}
+
+			startBlock := block.NumberU64() + ((uint64(startTimeUTC) - block.Header().Time) / averageBlockTime)
+			endBlock := block.NumberU64() + ((uint64(endTimeUTC) - block.Header().Time) / averageBlockTime)
```
lucca30: Risk of underflow if `uint64(startTimeUTC) < block.Header().Time`. Same for `endTimeUTC`

Should use something like:

```

if uint64(startTimeUTC) <= block.Header().Time  {
    return fmt.Errorf("start timestamp must be in the future (got %d <= %d)", startTimeUTC, now)
}

if uint64(endTimeUTC) <= block.Header().Time  {
    return fmt.Errorf("start timestamp must be in the future (got %d <= %d)", startTimeUTC, now)
}

```

x/bor/keeper/side_msg_server.go
```diff
@@ -181,6 +184,48 @@ func (s sideMsgServer) SideHandleMsgSpan(ctx sdk.Context, msgI sdk.Msg) sidetxs.
...
+
+	childBlock, err := s.k.contractCaller.GetBorChainBlock(ctx, nil)
+	if err != nil {
+		logger.Error("error fetching current child block", "error", err)
+		return sidetxs.Vote_VOTE_NO
+	}
+
+	childBlockNumber := childBlock.Number.Uint64()
+
+	if msg.DowntimeRange.StartBlock+types.PlannedDowntimeMinimumTimeInFuture <= childBlockNumber {
```
lucca30: If we want to ensure `msg.DowntimeRange.StartBlock` is at least `types.PlannedDowntimeMinimumTimeInFuture` in the future, we should move this constant to the right side on the failure check. Also change the operator to `<`:

```
if msg.DowntimeRange.StartBlock < childBlockNumber+types.PlannedDowntimeMinimumTimeInFuture
```

Also maybe make the message cleared:

```
logger.Error("start block must be at least minFuture in the future",
            "startBlock", msg.DowntimeRange.StartBlock,
			"minimumStartBlock",  childBlockNumber+types.PlannedDowntimeMinimumTimeInFuture+1,
		)

```

x/bor/keeper/veblop.go
```diff
@@ -425,14 +426,55 @@ func (k *Keeper) SelectProducer(ctx context.Context, currentProducer uint64, can
...
+			k.Logger(ctx).Debug("Selecting next producer (not down for range)", "currentProducer", currentProducer, "selected", c, "startBlock", startBlock, "endBlock", endBlock)
+			return c, nil
+		}
+
+		k.Logger(ctx).Debug("Skipping candidate producer (down for range)", "candidate", c, "startBlock", startBlock, "endBlock", endBlock)
+	}
+
+	// Full cycle and everyone is down: return the first candidate as a fallback
+	k.Logger(ctx).Error("All candidates are down for the requested range; falling back to first candidate", "fallback", candidates[0], "startBlock", startBlock, "endBlock", endBlock)
+	return candidates[0], nil
```
lucca30: Would be good have it documented. Since we know now the `candidates[0]` is always the most critical one. Do you know if this list is always at same order?

x/bor/keeper/veblop.go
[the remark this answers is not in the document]
avalkov: If there is span rotation it will happen before execution of SetProducerDowntime. In PostHandler of SetProducerDowntime we will know about this newly generated span.

x/bor/keeper/side_msg_server.go
avalkov: Thanks, fixed.

x/bor/keeper/veblop.go
avalkov: The list is ordered but there is no guarantee that the actually first producer wont be filter out earlier as inactive. This logic here should never trigger as the SetValidatorDowntime doesnt allow for all validators to be down during same block range.

x/bor/client/cli/tx.go
avalkov: Fixed.

x/bor/client/cli/tx.go
avalkov: Mostly just copy from other cli commands, deleted it.

x/bor/keeper/msg_server.go
```diff
@@ -259,6 +260,67 @@ func (s msgServer) BackfillSpans(ctx context.Context, msg *types.MsgBackfillSpan
 	return &types.MsgBackfillSpansResponse{}, nil
 }
 
+func (s msgServer) SetProducerDowntime(ctx context.Context, msg *types.MsgSetProducerDowntime) (*types.MsgSetProducerDowntimeResponse, error) {
+	var err error
```
marcello33: Shall we validate the `MsgSetProducerDowntime` first? It contains an address we can verify even before calling the `util.FormatAddress`

x/bor/keeper/side_msg_server.go
```diff
@@ -181,6 +184,48 @@ func (s sideMsgServer) SideHandleMsgSpan(ctx sdk.Context, msgI sdk.Msg) sidetxs.
 	return sidetxs.Vote_VOTE_YES
 }
 
+func (s sideMsgServer) SideHandleSetProducerDowntime(ctx sdk.Context, msgI sdk.Msg) sidetxs.Vote {
+	var err error
+	start := time.Now()
+	defer recordBorMetric(api.SideHandleMsgSpanMethod, api.SideType, start, &err)
```
marcello33: Shouldn't this be `api.SideHandleMsgSetProducerDowntime`?

x/bor/keeper/veblop.go
```diff
@@ -237,22 +235,25 @@ func (k *Keeper) ClearLatestFailedProducer(ctx context.Context) error {
 
 // SelectNextSpanProducer selects the next producer for a new span.
 // It calculates candidate set, filters by active producers and selects one.
-func (k *Keeper) SelectNextSpanProducer(ctx context.Context, currentProducer uint64, activeValidatorIDs map[uint64]struct{}, producerSetLimit uint64) (uint64, error) {
+func (k *Keeper) SelectNextSpanProducer(ctx sdk.Context, currentProducer uint64, activeValidatorIDs map[uint64]struct{}, producerSetLimit, startBlock, endBlock uint64) (uint64, error) {
```
marcello33: maybe we should keep using the ctx and unwrap it inside the function as we do elsewhere

x/bor/keeper/veblop.go
```diff
@@ -408,15 +409,16 @@ func (k *Keeper) FilterByActiveProducerSet(ctx context.Context, candidates []uin
...
 }
 
-// SelectProducer selects a producer from the candidates list.
-// The selected candidate will be the next candidate to the current producer.
-// If the current producer is not in the candidate list, the first candidate in the list will be chosen.
-func (k *Keeper) SelectProducer(ctx context.Context, currentProducer uint64, candidates []uint64) (uint64, error) {
+// SelectProducer selects a producer from the candidates list, starting after the current producer,
+// wrapping around. It skips any candidate that is down for the given [startBlock, endBlock].
+// If all candidates are down (full cycle), it returns candidates[0].
+func (k *Keeper) SelectProducer(ctx sdk.Context, currentProducer uint64, candidates []uint64, startBlock, endBlock uint64) (uint64, error) {
```
marcello33: Same as before with context? Just double checking as I remember we had some issues with contexts switches in the past in function signatures of keepers' methods

x/bor/types/util.go
```diff
@@ -93,3 +93,11 @@ func CalcCurrentBorSpanId(latestBorBlock uint64, latestHeimdallSpan *Span) (uint
 
 	return spanId, nil
 }
+
+const (
+	// TODO: Move to params?
```
marcello33: Are we considering this? Is it something we want to make it available for the community to update (e.g. via gov proposal)

x/bor/keeper/grpc_query.go
```diff
@@ -250,6 +250,33 @@ func (q queryServer) GetProducerVotes(ctx context.Context, req *types.QueryProdu
...
 }
 
+func (q queryServer) GetProducerPlannedDowntime(ctx context.Context, req *types.QueryProducerPlannedDowntimeRequest) (*types.QueryProducerPlannedDowntimeResponse, error) {
+	var err error
+	start := time.Now()
+	defer recordBorQueryMetric(api.GetProducerPlannedDowntimeMethod, start, &err)
+
+	if req == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "empty request")
+	}
```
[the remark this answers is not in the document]
avalkov: I dont see issues if somebody passes producerId that doesnt exist in the set?

x/bor/keeper/msg_server.go
```diff
@@ -259,6 +260,67 @@ func (s msgServer) BackfillSpans(ctx context.Context, msg *types.MsgBackfillSpan
...
+	}
+
+	if msg.DowntimeRange.EndBlock-msg.DowntimeRange.StartBlock < types.PlannedDowntimeMinRange {
+		return nil, fmt.Errorf("time range must be at least %d blocks", types.PlannedDowntimeMinRange)
+	}
+
+	if msg.DowntimeRange.EndBlock-msg.DowntimeRange.StartBlock > types.PlannedDowntimeMaxRange {
+		return nil, fmt.Errorf("time range must be at most %d blocks", types.PlannedDowntimeMaxRange)
+	}
+
```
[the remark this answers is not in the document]
avalkov: Added.

x/bor/keeper/side_msg_server.go
```diff
@@ -181,6 +184,48 @@ func (s sideMsgServer) SideHandleMsgSpan(ctx sdk.Context, msgI sdk.Msg) sidetxs.
...
+
+	if msg.DowntimeRange.StartBlock < childBlockNumber+types.PlannedDowntimeMinimumTimeInFuture {
+		logger.Error("start block must be at least minFuture in the future",
+			"startBlock", msg.DowntimeRange.StartBlock,
+			"minimumStartBlock", childBlockNumber+types.PlannedDowntimeMinimumTimeInFuture+1,
+		)
+		return sidetxs.Vote_VOTE_NO
+	}
+
+	if childBlockNumber+types.PlannedDowntimeMaximumTimeInFuture <= msg.DowntimeRange.EndBlock {
```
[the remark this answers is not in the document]
avalkov: Fixed. We should be checking if StartBlock is too much into the future.

PR #492, opened 2025-11-03
Adding 4th producer

avalkov:
# Description

Added HF checks for introducing 4th producer.

### DIFF
--- a/app/abci.go
+++ b/app/abci.go
@@ -26,11 +26,6 @@ import (
 	stakeTypes "github.com/0xPolygon/heimdall-v2/x/stake/types"
 )
 
-const (
-	ChangeProducerThreshold = 10
-	SpanRotationBuffer      = 20
-)
-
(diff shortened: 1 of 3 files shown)


x/bor/keeper/veblop.go
```diff
@@ -12,14 +12,24 @@ import (
...
+	producerSetLimit    = uint64(3)
+	newProducerSetLimit = uint64(4)
+)
+
+func GetProducerSetLimit(ctx sdk.Context) uint64 {
+	if ctx.BlockHeight() >= helper.GetSetProducerDowntimeHeight() {
+		return newProducerSetLimit
+	}
+	return producerSetLimit
+}
```
cffls: Might be a good idea to move these to https://github.com/0xPolygon/heimdall-v2/blob/develop/helper/config.go#L449-L481, where other HF params are specified.

avalkov: Done.

cffls (approved):

PR #493, opened 2025-11-04
perf(milestone): optimize RPC calls in block fetching

kamuikatsurgi:
# Description

Optimizes RPC calls in milestone proposition generation by fetching only available blocks and eliminating duplicate header fetches.


go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/0xPolygon/heimdall-v2
 
 // Note: Change the go image version in Dockerfile if you change this.
-go 1.24.6
+go 1.24.9
```
kamuikatsurgi: Same.

PR #499, opened 2025-11-12
ABCI layer improvements

avalkov:
# Description

ABCI layer improvements

### DIFF
--- a/app/vote_ext_utils.go
+++ b/app/vote_ext_utils.go
@@ -279,7 +279,7 @@ func FilterVoteExtensions(ctx sdk.Context, reqHeight int64, extVoteInfo []abciTy
 		}
 
 		if !cmtPubKey.VerifySignature(extSignBytes, vote.ExtensionSignature) {
-			return nil, fmt.Errorf("failed to verify validator %X vote extension signature", valAddrStr)
+			return nil, fmt.Errorf("failed to verify validator %s vote extension signature", valAddrStr)
 		}
 
 		sumVPPerBlockHash[common.Bytes2Hex(voteExtension.BlockHash)] += validator.VotingPower


app/vote_ext_utils.go
```diff
@@ -166,6 +166,164 @@ func ValidateVoteExtensions(ctx sdk.Context, reqHeight int64, extVoteInfo []abci
...
+			logger.Error("invalid milestone proposition detected for validator", "validator", valAddrStr, "error", err)
+			continue
+		}
+
+		// Check for duplicate votes by the same validator
+		if _, found := seenValidators[valAddrStr]; found {
+			return nil, fmt.Errorf("duplicate vote detected from validator %s at height %d", valAddrStr, reqHeight)
+		}
+		// Add validator address to the map
+		seenValidators[valAddrStr] = struct{}{}
```
lucca30: We are marking a validator as "seen" before confirming if their vote is valid. In case of a first malformed followed by a good one, we are going to fail

app/vote_ext_utils.go
```diff
@@ -166,6 +166,164 @@ func ValidateVoteExtensions(ctx sdk.Context, reqHeight int64, extVoteInfo []abci
...
+
+	majorityVP := totalVotingPower * 2 / 3
+	var majorityBlockHash string
+	for sumVPBlockHash, vp := range sumVPPerBlockHash {
+		if vp > majorityVP {
+			sumVP = vp
+			majorityBlockHash = sumVPBlockHash
+			break
+		}
+	}
```
lucca30: Is it deterministic? Since it's a map i think there is a risk of the iteration being non-deterministic.

app/vote_ext_utils.go
```diff
@@ -166,6 +166,164 @@ func ValidateVoteExtensions(ctx sdk.Context, reqHeight int64, extVoteInfo []abci
...
+		}
+
+		marshalDelimitedFn := func(msg proto.Message) ([]byte, error) {
+			var buf bytes.Buffer
+			if _, err := protoio.NewDelimitedWriter(&buf).WriteMsg(msg); err != nil {
+				return nil, err
+			}
+
+			return buf.Bytes(), nil
+		}
```
lucca30: we should consider creating this func outside the loop

PR #501, opened 2025-11-19
Enable test for SetValidatorDowntime tx

avalkov:
# Description

Enable test for SetValidatorDowntime tx

### DIFF
--- a/.github/workflows/kurtosis-e2e.yml
+++ b/.github/workflows/kurtosis-e2e.yml
@@ -102,6 +102,12 @@ jobs:
 
       - name: Run smoke tests (Checkpoints and Milestones)
         run: bash ./pos-workflows/tests/kurtosis_smoke_test.sh
+      
+      - name: Run set producer downtime test
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/kurtosis-e2e.yml
```diff
@@ -102,6 +102,12 @@ jobs:
 
       - name: Run smoke tests (Checkpoints and Milestones)
         run: bash ./pos-workflows/tests/kurtosis_smoke_test.sh
+      
+      - name: Run set producer downtime test
+        run: |
+          cd tests/producer_planned_downtime
```
kamuikatsurgi: This would be `cd ./pos-workflows/tests/producer_planned_downtime`

kamuikatsurgi (approved):
noice, If there's a way to make this test faster, that'd be good. Six and a half minutes is too long. if there isn't, nw.

marcello33 (approved):

cffls (approved):

PR #532, opened 2026-01-19
helper(tx): add EIP-1559 dynamic gas pricing for L1 transactions

kamuikatsurgi:
# Description

Implemented changes to create transaction auth objects with EIP-1559 dynamic fee pricing. The function fetches current network conditions and calculates optimal gas parameters:

- Retrieve `baseFee` from the latest block and the suggested tip from the network
- Cap tip to configured maximum `MainChainGasTipCap`
- Calculate fee cap as `(baseFee * 2) + tipCap` for fluctuation buffer
- Cap fee cap to configured maximum `MainChainGasFeeCap`

Removed unused `MainChainGasLimit` config option.

### DIFF
--- a/helper/config.go
+++ b/helper/config.go
@@ -59,8 +59,9 @@ const (
 	ClerkPollIntervalFlag        = "clerk_poll_interval"
 	SpanPollIntervalFlag         = "span_poll_interval"
 	MilestonePollIntervalFlag    = "milestone_poll_interval"
-	MainChainGasLimitFlag        = "main_chain_gas_limit"
-	MainChainMaxGasPriceFlag     = "main_chain_max_gas_price"
+
+	MainChainGasFeeCapFlag = "main_chain_gas_fee_cap"
+	MainChainGasTipCapFlag = "main_chain_gas_tip_cap"
 
 	NoACKWaitTimeFlag = "no_ack_wait_time"
 	ChainFlag         = "chain"
@@ -98,9 +99,8 @@ const (
 	DefaultSHCheckpointAckInterval = 30 * time.Minute
 	DefaultSHMaxDepthDuration      = 24 * time.Hour
 
-	DefaultMainChainGasLimit = uint64(5000000)
-
-	DefaultMainChainMaxGasPrice = 400000000000 // 400 Gwei
+	DefaultMainChainGasFeeCap = 500000000000 // 500 Gwei
+	DefaultMainChainGasTipCap = 10000000000  // 10 Gwei
 
 	DefaultBorChainID      = "15001"
 	DefaultHeimdallChainID = "heimdall-15001"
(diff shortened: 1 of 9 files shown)


helper/tx.go
```diff
@@ -21,77 +21,121 @@ import (
 )
 
 func GenerateAuthObj(client *ethclient.Client, address common.Address, data []byte) (auth *bind.TransactOpts, err error) {
-	// generate call msg
+	ctx := context.Background()
```
marcello33, line 41: I know we don't have tests for this function, but given its criticality and in the context of increasing coverage, can you please write some unit tests?

marcello33: Smth like testing these flows should be enough
  - Normal EIP-1559 flow                                                                                                                                                                                                                
  - baseFee = nil scenario                                                                                                                                                                                                              
  - Fee cap exceeding configured maximum                                                                                                                                                                                                
  - Tip cap exceeding fee cap

helper/tx.go
```diff
@@ -21,77 +21,121 @@ import (
...
+	if err != nil {
+		Logger.Error("unable to fetch suggested gas tip cap", "error", err)
 		return
 	}
 
-	nonce, err := client.NonceAt(context.Background(), fromAddress, nil)
+	// Use the lower of suggested tip or configured tip cap to prioritize lower fees.
+	// The configured tip cap acts as a maximum we're willing to pay.
+	gasTipCap := suggestedTipCap
+	if gasTipCap.Cmp(big.NewInt(configGasTipCap)) > 0 {
```
marcello33, line 92: So during network congestion, if the suggested tip exceeds the configured cap (10 Gwei), could transactions be delayed?                                         
If that's the case, maybe we can consider logging a warning when capping occurs

helper/config.go
```diff
@@ -147,9 +147,8 @@ type CustomConfig struct {
 
 	AmqpURL string `mapstructure:"amqp_url"` // amqp url
 
-	MainChainGasLimit uint64 `mapstructure:"main_chain_gas_limit"` // gas-limit for main chain txs, e.g., submit checkpoint
-
-	MainChainMaxGasPrice int64 `mapstructure:"main_chain_max_gas_price"` // max-gas-price for main chain txs
+	MainChainGasFeeCap int64 `mapstructure:"main_chain_gas_fee_cap"` // max fee per gas for EIP-1559 txs (in wei)
+	MainChainGasTipCap int64 `mapstructure:"main_chain_gas_tip_cap"` // max priority fee per gas for EIP-1559 txs (in wei)
```
marcello33, line 151: Shall we validate at configs load time that `MainChainGasTipCap ≤ MainChainGasFeeCap`?

helper/tx.go
kamuikatsurgi: Yes, transactions could be delayed during high congestion if we cap the tip. However, this value is configurable via configuration files — 10 Gwei is the default value. Validators could set it higher if they want, but most of them wouldn't want to pay huge gas for submitting checkpoints, as Ethereum transactions can be costly at this point.

Added a warn log. Thanks!

helper/tx.go
kamuikatsurgi: Added some unit tests, thanks!

helper/config.go
kamuikatsurgi: Addressed, thanks!

marcello33 (approved):

pratikspatil024 (approved):

PR #567, opened 2026-03-24
x/bor: add target producer id in producer planned downtime

kamuikatsurgi:
# Description

- Allows a producer declaring planned downtime to optionally nominate a specific replacement via `target_producer_id` on `MsgSetProducerDowntime`
- If the target is invalid, inactive, or down for the range, selection falls through to round-robin - the feature is best-effort, not a guarantee
- Requires an HF


kamuikatsurgi:
@claude review for the PR diff, and check for the broader consensus issues across the codebase.

helper/config.go
```diff
@@ -485,6 +486,7 @@ func InitHeimdallConfigWith(homeDir string, heimdallConfigFileFromFlag string) {
 		disableValSetCheckHeight = 25723063
 		initialHeight = 24404501
 		producerDowntimeHeight = 34966593
+		targetProducerOverrideHeight = 0 // TBD
 	case MumbaiChain:
```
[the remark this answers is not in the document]
kamuikatsurgi: Will address when we decide the HF number.

x/bor/keeper/msg_server.go
```diff
@@ -325,6 +325,23 @@ func (srv msgServer) SetProducerDowntime(ctx context.Context, msg *types.MsgSetP
 		return nil, fmt.Errorf("producer with address %s and id %d is not in the current producer set", msg.Producer, producerId)
 	}
 
+	// Validate target producer id if it is not 0 and the block height is greater than the target producer override height.
```
marcello33: It seems like we are validating the target against the candidates here (`msg.TargetProducerId == producerId`), but in the side handler we validate by fetching the full validator object via `GetValidatorFromValID` and then leverage `util.FormatAddress(targetValidator.Signer) == util.FormatAddress(msg.Producer)`
So we basically validate via ID first then via address.
I am thinking about the following: if a validator changes their signer address (e.g., `SignerUpdateMsg`), these checks could diverge, right? Nodes running the `msg_server` path would accept a message that the `side_msg_server` path rejects (or viceversa), causing a consensus failure. Anyway, it's always a good idea to unify the validation logic between handler and side handler (maybe even extract to a shared helper)

x/bor/keeper/msg_server.go
```diff
@@ -325,6 +325,23 @@ func (srv msgServer) SetProducerDowntime(ctx context.Context, msg *types.MsgSetP
...
+	if msg.TargetProducerId != 0 && sdkCtx.BlockHeight() >= helper.GetTargetProducerOverrideHeight() {
```
marcello33: Does the `TargetProducerId` need to be validated against min-max values or is it handled by proto? Maybe we could fail with Invalid Argument and 4xx bad request at start

x/bor/keeper/veblop.go
```diff
@@ -264,6 +264,37 @@ func (k *Keeper) SelectNextSpanProducer(ctx sdk.Context, currentProducer uint64,
...
+					"targetProducerID", targetProducerID, "error", err)
+			} else if !isDown {
+				k.Logger(ctx).Info("Using target producer override",
+					"currentProducer", currentProducer, "targetProducer", targetProducerID)
+				return targetProducerID, nil
+			} else {
+				k.Logger(ctx).Warn("Target producer is down for range, falling through to round-robin",
+					"targetProducerID", targetProducerID)
+			}
+		} else {
```
marcello33: Should this also check if `targetProducerID == currentProducer`?

x/bor/keeper/veblop_test.go
```diff
@@ -522,3 +522,252 @@ func (s *KeeperTestSuite) TestAddNewVeBlopSpan() {
...
+		setupMocks()
+
+		active := map[uint64]struct{}{2: {}, 3: {}}
+		result, err := borKeeper.SelectNextSpanProducer(ctx, 1, active, 2, 1002, 2001, 999)
+		require.NoError(err)
+		require.Equal(uint64(2), result)
+	})
+
+	// Target equals current producer — SelectNextSpanProducer doesn't filter this;
+	// the side handler and msg_server reject self-targeting upstream.
```
marcello33: Uhm ok this also relates to my [previous comment](https://github.com/0xPolygon/heimdall-v2/pull/567/changes#r2982718510)
So we only rely on `msg_server.go` catching this case? Because I see `SelectNextSpanProducer` is also called elsewhere from `abci.go`, I see we are usng 0 there but just as a defense mechanism I would not allow this. For example, if any future caller passes a non-zero value without pre-validation, the invariant could break, no? Maybe worth adding a `targetProducerID != currentProducer` check in `SelectNextSpanProducer` as well.

app/abci.go
```diff
@@ -794,7 +794,7 @@ func (app *HeimdallApp) checkAndAddFutureSpan(ctx sdk.Context, majorityMilestone
 			return err
 		}
 
-		err = app.BorKeeper.AddNewVeBlopSpan(ctx, currentProducer, lastSpan.EndBlock+1, endBlock, lastSpan.BorChainId, supportingValidatorIDs, uint64(ctx.BlockHeight()))
+		err = app.BorKeeper.AddNewVeBlopSpan(ctx, currentProducer, lastSpan.EndBlock+1, endBlock, lastSpan.BorChainId, supportingValidatorIDs, uint64(ctx.BlockHeight()), 0)
```
marcello33: Shall we use a constant like `RoundRobinDefault` to self-document this extra param and prevent accidental misuses?

x/bor/keeper/side_msg_server.go
```diff
@@ -276,6 +278,42 @@ func (srv sideMsgServer) SideHandleSetProducerDowntime(ctx sdk.Context, msgI sdk
 		return sidetxs.Vote_VOTE_NO
 	}
 
+	if msg.TargetProducerId != 0 && ctx.BlockHeight() >= helper.GetTargetProducerOverrideHeight() {
```
marcello33: Shall we early reject any `TargetProducerId != 0` before HF activation? Maybe cleaner

x/bor/keeper/msg_server.go
kamuikatsurgi: Thanks, addressed in 17abc786.

x/bor/keeper/msg_server.go
kamuikatsurgi: Addressed in 17abc786.

x/bor/keeper/veblop.go
kamuikatsurgi: Addressed in 17abc786.

app/abci.go
kamuikatsurgi: 17abc786

x/bor/keeper/veblop_test.go
kamuikatsurgi: 17abc786

x/bor/keeper/side_msg_server.go
kamuikatsurgi: 17abc786

marcello33 (dismissed):

kamuikatsurgi:
@claude review

cffls:
Looks good overall. Could you test a case where a cluster is half upgraded before the HF, a heimdall node with new version send out a rotation message request with target candidate, and make sure the consensus won't halt? Last time when the propose down time message was introduced, we had to add a check [here](https://github.com/0xPolygon/heimdall-v2/blob/develop/app/app.go#L512-L522).

x/bor/keeper/side_msg_server.go
```diff
@@ -642,7 +692,9 @@ func (srv sideMsgServer) PostHandleSetProducerDowntime(ctx sdk.Context, msgI sdk
 		return nil
 	}
 
-	if err := srv.k.AddNewVeBlopSpan(ctx, validatorId, msg.DowntimeRange.StartBlock, msg.DowntimeRange.StartBlock+params.SpanDuration, lastSpan.BorChainId, latestActiveProducer, uint64(ctx.BlockHeight())); err != nil {
+	// TargetProducerId is height-gated inside SelectNextSpanProducer;
+	// below the fork height it is silently ignored and round-robin is used.
```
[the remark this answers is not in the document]
kamuikatsurgi: Fixed

x/bor/keeper/veblop_test.go
```diff
@@ -522,3 +524,319 @@ func (s *KeeperTestSuite) TestAddNewVeBlopSpan() {
...
+
+func (s *KeeperTestSuite) TestSelectNextSpanProducerWithTarget() {
+	require := s.Require()
+	ctx := s.ctx
+	borKeeper := s.borKeeper
+	stakeKeeper := s.stakeKeeper
+
+	s.milestoneKeeper.EXPECT().GetLastMilestone(ctx).Return(&milestoneTypes.Milestone{
+		EndBlock: 1000,
+	}, nil).AnyTimes()
```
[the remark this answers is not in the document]
kamuikatsurgi: Fixed

helper/config.go
```diff
@@ -503,6 +506,7 @@ func InitHeimdallConfigWith(homeDir string, heimdallConfigFileFromFlag string) {
 		disableValSetCheckHeight = 10618299
 		initialHeight = 8788501
 		producerDowntimeHeight = 20457139
+		targetProducerOverrideHeight = 0 // TBD
```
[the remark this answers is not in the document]
kamuikatsurgi: Already a known todo

kamuikatsurgi:
> Looks good overall. Could you test a case where a cluster is half upgraded before the HF, a heimdall node with new version send out a rotation message request with target candidate, and make sure the consensus won't halt? Last time when the propose down time message was introduced, we had to add a check [here](https://github.com/0xPolygon/heimdall-v2/blob/develop/app/app.go#L512-L522).

Good catch, I added checks there as well similar to what we had for `MsgSetProducerDowntime`.

marcello33 (dismissed):

marcello33 (approved):

PR #570, opened 2026-03-26
fix(logs): enable millisecond-precision timestamps in heimdall logs

lucca30:
## Problem

Heimdall log lines carry only second-precision timestamps, e.g.:

```
2026-03-25T21:07:40Z INF Milestone stored successfully endBlock=84676411 ...
```

This makes it impossible to accurately measure sub-second timing between milestone events without relying on the Datadog ingestion timestamp — which carries a variable delay and is not a reliable high-resolution event marker.

## Root cause

`zerolog.TimeFieldFormat` (a zerolog global) defaults to `time.RFC3339`, which has second-level precision. This global controls how the timestamp is **stored** in the zerolog event before the `ConsoleWriter` ever formats it. `TimeFormatOption(time.RFC3339Nano)` was already set for the ConsoleWriter display layer, but the stored value only carried second granularity — so the displayed timestamp was always truncated to the second regardless of display format.

## Fix

Set `zerolog.TimeFieldFormat` to a fixed 3-decimal-place millisecond format (`"2006-01-02T15:04:05.000Z07:00"`) before logger construction, and apply the same format to `TimeFormatOption`. This makes all log lines emit millisecond-precision timestamps that are consistent with bor's log format:

```
2026-03-26T13:54:28.593Z INF Generated a new milestone proposition ...
2026-03-26T13:54:28.793Z INF 2/3rd majority reached on milestone proposition ...
2026-03-26T13:54:28.793Z INF Milestone stored successfully ...
```

## Motivation

The change is driven by block-finality latency analysis. The key heimdall events being timed are:

- `Generated a new milestone proposition` — proposal phase (T2)
- `2/3rd majority reached on milestone proposition` — consensus achieved (T3)
- `Milestone stored successfully` — finality committed (T4)

With second precision, derived latencies have ±500ms noise per step. With millisecond precision, inter-event timing becomes accurate enough to identify the dominant bottleneck in the finality pipeline (CometBFT consensus round duration, validator nil-vote behavior) without relying on Datadog ingestion timestamps.

Milliseconds (3 decimal places) were chosen over nanoseconds to match bor's existing log timestamp format, keeping both services consistent for cross-service log analysis.

## Validation

Tested on a local kurtosis devnet — confirmed all three milestone log events carry millisecond-precision timestamps after the fix.

### DIFF
--- a/cmd/heimdalld/cmd/root.go
+++ b/cmd/heimdalld/cmd/root.go
@@ -2,7 +2,6 @@ package heimdalld
 
 import (
 	"os"
-	"time"
 
 	"cosmossdk.io/log"
 	"github.com/cometbft/cometbft/cmd/cometbft/commands"
@@ -137,11 +136,12 @@ func NewRootCmd() *cobra.Command {
 			}
 
 			logNoColor := serverCtx.Viper.GetBool(flags.FlagLogNoColor)
-			// Store timestamps with sub-second precision so that log lines carry
-			// millisecond granularity. Without this, zerolog stores the Unix
-			// timestamp as RFC3339 (second precision) before the ConsoleWriter
-			// ever gets a chance to format it, resulting in second-only output.
-			zerolog.TimeFieldFormat = time.RFC3339Nano
+			// Store timestamps with millisecond precision so that log lines carry
+			// sub-second granularity. Without this, zerolog stores the timestamp
+			// as RFC3339 (second precision) before the ConsoleWriter ever gets a
+			// chance to format it, resulting in second-only output.
+			// Using 3 decimal places ("000") matches bor's log timestamp format.
+			zerolog.TimeFieldFormat = "2006-01-02T15:04:05.000Z07:00"
 			var logOpts []log.Option
 			if serverCtx.Viper.GetString(flags.FlagLogFormat) == flags.OutputFormatJSON {
 				logOpts = append(logOpts, log.OutputJSONOption())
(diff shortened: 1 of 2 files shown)


cmd/heimdalld/cmd/root.go
```diff
@@ -145,7 +150,7 @@ func NewRootCmd() *cobra.Command {
 			}
 			logOpts = append(logOpts,
 				log.LevelOption(logLevel),
-				log.TimeFormatOption(time.RFC3339Nano),
+				log.TimeFormatOption("2006-01-02T15:04:05.000Z07:00"),
```
marcello33: maybe we can use a constant here, since it's duplicated from line 144

helper/config.go
```diff
@@ -368,7 +368,7 @@ func InitHeimdallConfigWith(homeDir string, heimdallConfigFileFromFlag string) {
 	}
 	logOpts = append(logOpts,
 		logger.LevelOption(logLevel),
-		logger.TimeFormatOption(time.RFC3339Nano),
+		logger.TimeFormatOption("2006-01-02T15:04:05.000Z07:00"),
```
marcello33: Same here

marcello33 (dismissed):
Small nit, LGTM else

cmd/heimdalld/cmd/root.go
lucca30: Good call, extracted to `helper.LogTimestampFormat` — updated all three sites in root.go and helper/config.go.

helper/config.go
lucca30: Same here — uses `helper.LogTimestampFormat` now.

manav2401 (approved):

marcello33 (approved):

PR #604, opened 2026-06-04
build: bump Go to 1.26.4, add Dependabot, govulncheck nightly

kamuikatsurgi:
## Summary

Three changes to reduce govulncheck noise on PRs.

### 1. Bump Go 1.26.3 → 1.26.4 (fixes current govuln findings)

Clears two stdlib advisories that govulncheck has been flagging on every PR:
- **GO-2026-5039**: Arbitrary inputs in errors without escaping in `net/textproto` (called via `util.GetUnconfirmedTxnCount` → `io.ReadAll` → `textproto.Reader.ReadMIMEHeader`)
- **GO-2026-5037**: Inefficient candidate hostname parsing in `crypto/x509` (called via `queue.NewQueueConnector` → `amqp091.Dial`, `helper.InitHeimdallConfigWith` → `rpc.Dial`, and the bor side-msg server)

After the bump, `govulncheck ./...` reports `No vulnerabilities found.` locally.

### 2. Move `govulncheck` workflow off PRs onto a nightly schedule

The workflow now runs daily at `00:00 UTC` (plus `workflow_dispatch`) instead of on every `push` / `pull_request`. PR runs have been failing too often because they track upstream Go stdlib advisories that arrive between Go patch releases — creating noise on unrelated PRs and blocking merges on issues unrelated to the change. The nightly job still surfaces new advisories.

### 3. Add Dependabot config

Adds `.github/dependabot.yml` to automate the recurring third-party dep bumps (`golang.org/x/net`, `golang.org/x/crypto`, etc.) that govulncheck has historically required us to do by hand. Covers `gomod`, `docker`, and `github-actions`. Runs **daily**, groups minor/patch updates to limit PR volume, and **ignores 0xPolygon-maintained forks** (cosmos-sdk, cometbft, bor) which must be pinned by hand for consensus determinism per the security rules.

Caveat: Dependabot does **not** bump the Go `toolchain` directive itself, so stdlib CVEs like the two fixed in this PR will still need manual Go bumps surfaced by the nightly govulncheck job.

## Files touched
- `go.mod` — `go 1.26.4`
- `.golangci.yml` — `go: '1.26.4'`
- `.github/workflows/govulncheck.yml` — schedule (`0 0 * * *`) + `workflow_dispatch`; drop `get-diff-action`, `push`, and `pull_request` triggers
- `.github/dependabot.yml` — new

Note: `Dockerfile` already uses the floating `golang:1.26-alpine` tag, which picks up 1.26.4 automatically — no Dockerfile change needed in this PR.

## Test plan
- [ ] CI passes on this PR (govulncheck no longer runs on PRs, so its previous failure does not gate this)
- [ ] Manually trigger the new `Govuln` workflow via `workflow_dispatch` on `develop` after merge and confirm it reports clean
- [ ] Confirm Dependabot starts producing PRs within ~24h of merge

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,42 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    open-pull-requests-limit: 10
+    commit-message:
+      prefix: "build"
+    groups:
+      go-deps:
+        update-types:
+          - "minor"
+          - "patch"
+    ignore:
+      # 0xPolygon-maintained forks pinned via replace directives for
+      # consensus determinism. Bump these by hand.
+      - dependency-name: "github.com/0xPolygon/cosmos-sdk*"
+      - dependency-name: "github.com/0xPolygon/cometbft*"
+      - dependency-name: "github.com/0xPolygon/bor*"
+
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    commit-message:
+      prefix: "build"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    commit-message:
+      prefix: "ci"
+    groups:
+      actions:
+        update-types:
+          - "minor"
+          - "patch"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,8 +3,9 @@ updates:
   - package-ecosystem: 'gomod'
     directory: '/'
     schedule:
-      interval: 'daily'
-    open-pull-requests-limit: 10
+      interval: 'weekly'
+      day: 'monday'
+    open-pull-requests-limit: 3
     commit-message:
       prefix: 'build'
     groups:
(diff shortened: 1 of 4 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,42 @@
...
+      day: 'monday'
+    commit-message:
+      prefix: 'build'
+
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
+      interval: 'weekly'
+      day: 'monday'
+    commit-message:
```
marcello33, line 36: Is this resolving the open PRs from dependabot and merging them too? If that's the case, will it run any sort of e2e tests before merging? We had issues in the past where dependabot would actually break things..

kamuikatsurgi: This would just open the PRs, not merge or commit. If all the e2e tests pass on the PR, we can merge, otherwise not.

marcello33: Ok that commit-message confused, me. Thanks LGTM

marcello33 (approved):

marcello33:
@kamuikatsurgi we have nightly runs for race tests in bor, and they report to slack channel if failing. Can we do the same here?

lucca30 (approved):

PR #618, opened 2026-06-30
bor: fall back to eligible producers when veblop candidate set is empty

pratikspatil024:
## Summary

veBlop span producer selection (`SelectNextSpanProducer`) could hand `SelectProducer` an empty candidate slice when the elected producer set collapsed to only the current producer and that producer was filtered out of the active/supporting set. In the future-span path the resulting error propagates out of `PreBlocker` before the milestone/span commit, so block finalization cannot make progress — a deterministic halt with no self-heal.

This adds a deterministic, **Ithaca-gated** fallback that fires **only on the future-span path**. `SelectNextSpanProducer` / `AddNewVeBlopSpan` take an opt-in `fallbackToActiveSet` flag, and only `checkAndAddFutureSpan` (the fatal `PreBlocker` path) passes it. When that path's candidate list would otherwise be empty, the fallback draws a replacement from the caller's active set (the milestone supporters), then from the full validator set as defense-in-depth, each sorted ascending and excluding the current and excluded producers. In the future-span path this is always non-empty — a >2/3 milestone that excludes the current producer leaves a non-empty supporting set — so producer selection no longer errors there.

The non-fatal rotation / pending-stall / side-message paths **do not opt in** and are unchanged: post-Ithaca they still hit `SelectProducer([])` → `"no candidates found"` → skip-and-retry, exactly as before. Scoping the fallback to the future-span path (rather than sharing it across all callers) keeps the rotation paths' self-heal contract intact and avoids reinstating a known-failed producer there.

The fallback helpers live in `x/bor/keeper/veblop_producer_fallback.go` (`eligibleProducerFallback`, `sortedEligibleProducers`) to keep `veblop.go` within size/complexity budgets.

## Executed tests

- **Unit (`x/bor/keeper`)**: `eligibleProducerFallback` branch coverage (active-set preference, validator-set fallback, excluded-set handling, validator-set read error → nil), `sortedEligibleProducers`, and `SelectNextSpanProducer`'s empty-candidate path — pre-Ithaca still errors; post-Ithaca **without** opt-in still errors (rotation-path behavior preserved); post-Ithaca **with** opt-in falls back to a non-current producer.
- **App-level regression (`app`)**: `PreBlocker` future-span creation when the elected set has collapsed to the current producer and it withholds its milestone vote — pre-Ithaca still errors (`no candidates found`), post-Ithaca commits the milestone + span with a milestone-supporting validator as the new producer.
- Full `x/bor/...` and `app/...` suites, `go vet`, `gofmt` clean.
- **diffguard** diff-mode incl. mutation testing: passing (Tier-1 logic mutation kill 100%, Tier-2 100%).
- **Kurtosis devnet**: confirmed the fallback fires on a running network — the empty-candidate selection error no longer occurs and spans are created with the fallback producer. Note: a long-running devnet in which every validator votes for a single producer is *intentionally pathological* (a permanent single-elected-producer state oscillates regardless of this change) and is **not required for verification**; the deterministic unit/app tests above cover the behavior.

## Rollout notes

- **Consensus-affecting**: changes deterministic producer selection in the degenerate empty-candidate case, on the future-span path only. **Gated behind the Ithaca hardfork height**, which is unset (`0`) on mainnet, Amoy, and local today, so behavior is unchanged until the coordinated Ithaca activation — it ships with that upgrade alongside the related forced-span-rotation work.
- No state schema, store, migration, proto, or genesis changes. Backwards-compatible until the fork height.

### DIFF
--- a/app/veblop_producer_fallback_test.go
+++ b/app/veblop_producer_fallback_test.go
@@ -1,0 +1,217 @@
+package app
+
+import (
+	"strings"
+	"testing"
+
+	abci "github.com/cometbft/cometbft/abci/types"
+	cmtsecp256k1 "github.com/cometbft/cometbft/crypto/secp256k1"
+	cmtproto "github.com/cometbft/cometbft/proto/tendermint/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/require"
+
+	"github.com/0xPolygon/heimdall-v2/helper"
+	"github.com/0xPolygon/heimdall-v2/sidetxs"
+	borKeeper "github.com/0xPolygon/heimdall-v2/x/bor/keeper"
+	borTypes "github.com/0xPolygon/heimdall-v2/x/bor/types"
+	milestoneTypes "github.com/0xPolygon/heimdall-v2/x/milestone/types"
+	stakeTypes "github.com/0xPolygon/heimdall-v2/x/stake/types"
+)
+
+// singletonScenario holds the staged degenerate VEBLOP state: the elected producer set
+// has collapsed to the single current producer, and a >2/3 milestone is finalized by the
+// other validators while the current producer withholds its milestone vote (so it is
+// excluded from the supporting set). Future-span creation in PreBlocker then asks
+// SelectNextSpanProducer for a replacement and the candidate set (singleton minus the
+// current producer) is empty.
+type singletonScenario struct {
+	app               *HeimdallApp
+	ctx               sdk.Context
+	height            int64
+	lastSpan          borTypes.Span
+	currentProducerID uint64
+	validators        []*stakeTypes.Validator
+	extCommitBytes    []byte
+}
+
+func stageSingletonProducer(t *testing.T) *singletonScenario {
+	t.Helper()
+	_, app, ctx, validatorPrivKeys := SetupAppWithABCICtxAndValidators(t, 4)
+
+	oldRio := helper.GetRioHeight()
+	helper.SetRioHeight(100)
+	t.Cleanup(func() { helper.SetRioHeight(oldRio) })
+
+	height := app.LastBlockHeight() + 1
+	ctx = ctx.WithBlockHeight(height)
+
+	validators := app.StakeKeeper.GetAllValidators(ctx)
+	require.Len(t, validators, 4)
+
+	currentProducerID := validators[0].ValId
+	lastSpan := borTypes.Span{
+		Id:                1,
+		StartBlock:        100,
+		EndBlock:          199,
+		BorChainId:        "1",
+		ValidatorSet:      stakeTypes.ValidatorSet{Validators: validators, Proposer: validators[0]},
+		SelectedProducers: []stakeTypes.Validator{*validators[0]},
+	}
+	require.NoError(t, app.BorKeeper.AddNewSpan(ctx, &lastSpan))
+
+	// Every validator casts a one-entry ballot for the current producer, collapsing
+	// CalculateProducerSet to the singleton [currentProducer].
+	msgServer := borKeeper.NewMsgServerImpl(app.BorKeeper)
+	for _, validator := range validators {
+		_, err := msgServer.VoteProducers(ctx, &borTypes.MsgVoteProducers{
+			Voter:   validator.Signer,
+			VoterId: validator.ValId,
+			Votes:   borTypes.ProducerVotes{Votes: []uint64{currentProducerID}},
+		})
+		require.NoError(t, err)
+	}
+	candidates, err := app.BorKeeper.CalculateProducerSet(ctx, helper.GetProducerSetLimit(ctx))
+	require.NoError(t, err)
+	require.Equal(t, []uint64{currentProducerID}, candidates)
+
+	previousMilestoneHash := common.HexToHash("0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa").Bytes()
+	require.NoError(t, app.MilestoneKeeper.AddMilestone(ctx, milestoneTypes.Milestone{
+		Proposer:        validators[1].Signer,
+		Hash:            previousMilestoneHash,
+		StartBlock:      1,
+		EndBlock:        lastSpan.StartBlock - 1,
+		BorChainId:      "1",
+		MilestoneId:     "previous",
+		Timestamp:       1,
+		TotalDifficulty: 1,
+	}))
+
+	winningMilestone := &milestoneTypes.MilestoneProposition{
+		StartBlockNumber: lastSpan.StartBlock,
+		ParentHash:       previousMilestoneHash,
+		BlockHashes:      [][]byte{common.HexToHash("0xbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb").Bytes()},
+		BlockTds:         []uint64{2},
+	}
+
+	// Validators 1, 2 and 3 (75% of voting power) finalize the milestone; the current
+	// producer (validator 0) withholds, so it is absent from supportingValidatorIDs.
+	extCommitBytes := milestoneExtendedCommit(t, app.ChainID(), height, validators, validatorPrivKeys, currentProducerID, winningMilestone)
+
+	return &singletonScenario{
+		app:               app,
+		ctx:               ctx,
+		height:            height,
+		lastSpan:          lastSpan,
+		currentProducerID: currentProducerID,
+		validators:        validators,
+		extCommitBytes:    extCommitBytes,
+	}
+}
+
+// Pre-Ithaca the fallback is gated off, so the singleton-minus-current empty set still
+// errors out of PreBlocker. This pins the gate boundary: the fix changes nothing before
+// the fork activates.
+func TestVeBlopEmptyElectedSetPreIthacaErrors(t *testing.T) {
+	s := stageSingletonProducer(t)
+
+	oldIthaca := helper.GetIthacaHeight()
+	helper.SetIthacaHeight(0) // disabled: IsIthaca is always false
+	t.Cleanup(func() { helper.SetIthacaHeight(oldIthaca) })
+
+	_, err := s.app.PreBlocker(s.ctx, &abci.RequestFinalizeBlock{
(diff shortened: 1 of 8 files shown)


pratikspatil024:
@claude review once

x/bor/keeper/veblop_producer_fallback.go
```diff
@@ -0,0 +1,57 @@
...
+		k.Logger(ctx).Error("Failed to get validator set for producer fallback", "error", err)
+		return nil
+	}
+
+	valIDs := make(map[uint64]struct{}, len(valSet.Validators))
+	for _, v := range valSet.Validators {
+		valIDs[v.ValId] = struct{}{}
+	}
+	return sortedEligibleProducers(valIDs, currentProducer, excludedProducerIDs)
+}
```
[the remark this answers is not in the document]
pratikspatil024: Addressed by scoping the fallback to the future-span path only, which removes the rotation path from this scenario entirely.

The empty-candidate recovery is now an explicit opt-in: `SelectNextSpanProducer` / `AddNewVeBlopSpan` take a `fallbackToActiveSet` flag, and **only** `checkAndAddFutureSpan` (the fatal `PreBlocker` path) passes `true`. `checkAndRotateCurrentSpan`, `rotateSpanFromPendingHead`, and the side-message path don't opt in, so post-Ithaca they still hit `SelectProducer([])` → `"no candidates found"` → `AddNewVeBlopSpan` errors → skip-and-retry, exactly as pre-Ithaca. `eligibleProducerFallback` is unreachable from those callers, so the validator-set branch can no longer reinstate a known-failed producer — your step 4 onward can't be entered for the rotation path.

I chose to remove the fallback from the rotation path rather than thread `{currentProducer} ∪ latestFailedProducer` through as `excludedProducerIDs`, since the intent is to fix only the future-span halt; the rotation / pending-stall paths already self-heal via skip-and-retry. Also updated the `eligibleProducerFallback` docstring to drop the rotation-path language and the PR description to match.

Done in b17b319f. diffguard diff-mode (mutation) green — T1 100%, T2 100%; full `x/bor/keeper` and `app` suites pass.

app/veblop_producer_fallback_test.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+func stageSingletonProducer(t *testing.T) *singletonScenario {
+	t.Helper()
+	_, app, ctx, validatorPrivKeys := SetupAppWithABCICtxAndValidators(t, 4)
+
+	oldRio := helper.GetRioHeight()
+	helper.SetRioHeight(100)
+	t.Cleanup(func() { helper.SetRioHeight(oldRio) })
+
```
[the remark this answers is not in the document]
pratikspatil024: Verified safe as-is, and it matches the established convention in this package.

`rioHeight` / `ithacaHeight` are package-level globals in `helper`, and every fork-height-mutating test in `app/` is sequential (no `t.Parallel()`) — `abci_test.go`, `pending_stall_test.go`, `abci_additional_test.go`, `ante_test.go`, `vote_ext_utils_test.go`, and this file (~60 call sites). Go runs all sequential tests to completion (including their `t.Cleanup` restores) before resuming any `t.Parallel()` test, so a fork-height writer never runs concurrently with another test. The only two parallel tests in the package — `TestValidateGenesis` and `TestGetMaccPerms` (app_test.go:179, :197) — neither read nor write fork heights, so the parallel phase has no writer at all.

`go test ./app/ -race -count=1` passes clean on the full package.

Making fork heights test-local would require the production getters (`IsIthaca`, `GetRioHeight`), which run on hot consensus paths, to stop reading globals — out of scope here and inconsistent with the rest of the suite. Keeping this test sequential, consistent with the package convention.

pratikspatil024:
@claude review once

x/bor/keeper/veblop.go
```diff
@@ -268,6 +273,10 @@ func (k *Keeper) SelectNextSpanProducer(ctx sdk.Context, currentProducer uint64,
 		activeCandidates = filterExcludedProducers(newCandidates, excludedProducerIDs)
 	}
 
+	// Post-Ithaca, recover a candidate so the future-span path (the only opt-in via fallbackToActiveSet) can't hand SelectProducer an empty set (see eligibleProducerFallback).
+	if len(activeCandidates) == 0 && len(fallbackToActiveSet) > 0 && fallbackToActiveSet[0] && helper.IsIthaca(ctx.BlockHeight()) {
+		activeCandidates = k.eligibleProducerFallback(ctx, currentProducer, activeValidatorIDs, excludedProducerIDs)
+	}
```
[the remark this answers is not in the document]
pratikspatil024: Good observation, but I'm leaving this as-is — it's diagnostic-only and the empty-fallback case is effectively unreachable on the only opted-in path.

`checkAndAddFutureSpan` is the sole caller that opts in, and it passes `activeValidatorIDs = supportingValidatorIDs` with `excludedProducerIDs = nil`. So:

- Step 1 of `eligibleProducerFallback` (`sortedEligibleProducers(supporters, currentProducer, nil)`) is non-empty whenever there is ≥1 milestone supporter — and we only reach this path because a >2/3 milestone that excludes the current producer finalized, so the supporting set is non-empty by construction.
- The "everything excluded" case can't occur here, since `excludedProducerIDs` is `nil`.
- The only remaining ways to get an empty slice are a `GetValidatorSet` read error (already logged inside `eligibleProducerFallback`) or a single-validator network — both pathological.

A more specific error also wouldn't change behavior: empty → error out of `PreBlocker` → halt, regardless of the message. Against that, `SelectNextSpanProducer` is already at its complexity/size budget (diffguard flags it), so adding a branch here would trade real complexity for a message that essentially never fires. Keeping the existing `"no candidates found"` path; the `GetValidatorSet` failure is the diagnostic signal that actually matters and it's already logged.

x/bor/keeper/veblop.go
```diff
@@ -14,12 +14,17 @@ import (
...
 
-// AddNewVeBlopSpan adds a new veBlop (Validator-elected block producer) span
-func (k *Keeper) AddNewVeBlopSpan(ctx sdk.Context, currentProducer uint64, startBlock uint64, endBlock uint64, borChainID string, activeValidatorIDs map[uint64]struct{}, heimdallBlock uint64, targetProducerID uint64, excludedProducerIDs map[uint64]struct{}) error {
+// AddNewVeBlopSpan adds a new veBlop (Validator-elected block producer) span.
+//
+// fallbackToActiveSet opts the caller into the post-Ithaca empty-candidate recovery in
+// SelectNextSpanProducer. Only the future-span PreBlocker path passes true: there an empty
+// candidate set is a fatal chain halt. The rotation / pending-stall / side-message callers omit
+// it (default false) and keep their existing skip-and-retry behavior on an empty selection.
+func (k *Keeper) AddNewVeBlopSpan(ctx sdk.Context, currentProducer uint64, startBlock uint64, endBlock uint64, borChainID string, activeValidatorIDs map[uint64]struct{}, heimdallBlock uint64, targetProducerID uint64, excludedProducerIDs map[uint64]struct{}, fallbackToActiveSet ...bool) error {
```
cffls: This function has grown quite large. Is it possible to simplify and reduce the number of parameters? Also, this `...bool` does need to be a variadic parameter I guess?

pratikspatil024: Hi @cffls - addressed the variadic part ([here](https://github.com/0xPolygon/heimdall-v2/pull/618/changes/55658b8fb99aa499583457a1d7f10ebe32e1bd7a)) by making `fallbackToActiveSet` a plain `bool` and keeping the future-span path as the only caller that passes `true`.

On the size/parameter count, I agree.  But it's largely pre-existing. `SelectNextSpanProducer` / `AddNewVeBlopSpan` were already at 8/9 parameters and high complexity before this PR; this change only adds the one fallbackToActiveSet parameter and one branch.
I did try the obvious cleanup, but `AddNewVeBlopSpan` is shared by the rotation / pending-stall / side-message paths, so fixing that pulls a lot of pre-existing, unrelated code into diffguard's mutation/size scope (mutation fell to ~73%, several size violations), and cleaning that back up would mean adding a batch of tests for pre-existing branches.

Do you think I should address that in this PR or do it as a standalone PR once this is merged?

cffls: Yeah, agreed it is already too complex before this PR. I think we can keep it as it is now and maybe refactor in the future.

cffls (approved):

adamdossa:
@claude review once

app/veblop_producer_fallback_test.go
```diff
@@ -0,0 +1,217 @@
...
+	borTypes "github.com/0xPolygon/heimdall-v2/x/bor/types"
+	milestoneTypes "github.com/0xPolygon/heimdall-v2/x/milestone/types"
+	stakeTypes "github.com/0xPolygon/heimdall-v2/x/stake/types"
+)
+
+// singletonScenario holds the staged degenerate VEBLOP state: the elected producer set
+// has collapsed to the single current producer, and a >2/3 milestone is finalized by the
+// other validators while the current producer withholds its milestone vote (so it is
+// excluded from the supporting set). Future-span creation in PreBlocker then asks
+// SelectNextSpanProducer for a replacement and the candidate set (singleton minus the
```
marcello33, line 26: is this ok for operators running M:N Bor:Heimdall and some running Heimdall with no connected Bor? Wondering if it can freeze a span for a validator that produces no Bor blocks, turning recovery into another stall.

pratikspatil024: Good question. One clarification first: milestone support means the validator's Bor is **observing** the chain (it read the tip to build the proposition) — that's not the same as being able to **produce** blocks as the elected producer, so the fallback favors live, chain-tracking validators but doesn't guarantee the one it picks will seal blocks.

That said, this does not create a new Heimdall consensus risk for M:N or Heimdall-only operators:

- **No new hard Heimdall→Bor dependency**: selection is deterministic from consensus inputs (producer votes, validator set, and the finalized milestone-supporting vote extensions), so Heimdall-only nodes compute the same result and stay in consensus. A Heimdall with no reachable Bor should also not normally enter the fallback's first-choice supporter set, because it cannot generate a milestone proposition.
- **On "another stall"**: if the fallback picks an observe-only validator, the new span can still stall at Bor production. But that is no longer a fatal `PreBlocker` halt; Heimdall remains live and the existing POS-3629 pending-stall/rotation paths can attempt recovery/retry. If no selected candidate can actually produce, that is a broader producer-capability/liveness limit independent of this change.
- All gated behind Ithaca, so no change until activation.

marcello33 (approved):

PR #621, opened 2026-07-10
app: cap messages per tx in CheckTx

marcello33:
## Summary

Adds a per-transaction message-count bound enforced at `CheckTx` (mempool admission). A single transaction may now carry at most `maxMsgsPerTx = 16` top-level messages; a transaction exceeding that is rejected with
`ErrTxTooLarge` before the per-message decode/validation loop runs.

In normal operation every Heimdall transaction carries exactly **one** top-level message — all submission paths (bridge broadcaster and CLI) build single-message txs, and there is no batching anywhere, including the
clerk/state-sync path. The cap of 16 is therefore 16× the real maximum and cannot reject legitimate traffic; it matches the existing `maxMultiSendOutputs = 16` bound already used in the ante stack.

Without this bound, a single large multi-message transaction forces every node to do O(number-of-messages) decode and validation work on each gossip hop before the transaction is ultimately rejected downstream. Capping the
message count at admission removes that amplification cheaply.

Note: the existing `SideTxDecorator` only bounds the number of *side* messages per tx (≤1), not the total message count — so this guard is additive and is the right place to bound `len(tx.GetMsgs())` directly.

## Executed tests

- Unit test (`app/app_test.go`): drives real encoded txs through `CheckTx` with counts 1 / 16 / 17 / 1600 (+ a recheck-path case), asserting the boundary: `>16` → `ErrTxTooLarge`, `≤16` → passes the cap into the ante chain.
- Diffguard (with mutation): mutation 100% (3/3), cognitive-complexity / dependency / dead-code all clean. The only flag is `CheckTx` function length (57 lines vs the 50 threshold) — pre-existing (base was already 51; this change adds 6). Left as-is to avoid churning the consensus-adjacent veBlop loop in `CheckTx`; a follow-up can decompose that function.
- Kurtosis small devnet (built from this branch), confirmed end-to-end:
  - A 2000-message and a 17-message tx are rejected at `CheckTx` with `ErrTxTooLarge` ("exceeds per-tx limit of 16"); a 16-message tx passes the cap and proceeds to the ante chain (fails later at signature check, not the cap). Boundary is exact and fires before decode.
  - Regression: state syncs (clerk `MsgEventRecord`), checkpoints, and spans all commit normally with the cap in place — no legitimate flow is affected.
  - Under a burst of oversized txs, heimdall CPU stayed flat and the chain kept producing blocks (rejected pre-decode → no processing cost).

## Rollout notes

- **Not consensus-affecting, no hardfork required.** The bound is enforced only in the overridden `CheckTx` (mempool admission + gossip), not in `DeliverTx` / `ProcessProposal` / block execution. Block validity and the app hash are unchanged, so no coordinated upgrade is needed.
- **Backwards-compatible.** No proto, state, or API changes; legitimate single-message traffic is unaffected.
- Beneficial to prioritise on sentries / RPC nodes (they front the public API and gossip mempool traffic), but it can roll out incrementally.
- Follow-ups tracked separately (optional): rejection observability (a Prometheus counter and/or a rate-limited log) and a hardfork-gated `ValidateBasic` for the affected message type.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -103,6 +103,12 @@ const (
 	// HTTP header constants
 	headerContentType       = "Content-Type"
 	mimeTypeApplicationJSON = "application/json"
+
+	// maxMsgsPerTx bounds how many messages a single tx may carry through CheckTx.
+	// Capping here avoids mempool amplification where one oversized tx packs thousands of messages,
+	// forcing decode and validation on every gossip hop before rejection.
+	// Enforced in CheckTx only, so block validity and consensus are unaffected.
+	maxMsgsPerTx = 16
 )
 
 var (
@@ -496,6 +502,12 @@ func (app *HeimdallApp) CheckTx(req *abci.RequestCheckTx) (*abci.ResponseCheckTx
 		}
 
 		msgs := tx.GetMsgs()
+		if len(msgs) > maxMsgsPerTx {
+			return &abci.ResponseCheckTx{
+				Code: sdkerrors.ErrTxTooLarge.ABCICode(),
+				Log:  fmt.Sprintf("tx carries %d messages, exceeds per-tx limit of %d", len(msgs), maxMsgsPerTx),
+			}, nil
+		}
 		for _, msg := range msgs {
 			// Check for MsgVoteProducers and apply VEBLOP validation
 			if _, ok := msg.(*borTypes.MsgVoteProducers); ok {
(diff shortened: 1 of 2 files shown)


marcello33:
@claude review once

pratikspatil024 (dismissed):

kamuikatsurgi (dismissed):

marcello33:
2 approvals got discarded by a new commit which was simply fixing a comment. Merged bypassing the checks.

PR #634, opened 2026-08-08
fix(stake): guard nil amount before IsZero/IsNegative in ValidateBasic

SashaMIT:
## Summary
- `MsgValidatorJoin` / `MsgStakeUpdate` `ValidateBasic` called `IsZero()` / `IsNegative()` on possibly nil `math.Int` (proto field omitted).
- Nil Int panics on those methods. `ValidateBasic` runs ahead of signature verification in the ante handler, so the panic is reachable unauthenticated.
- Same class as the existing topup `Fee` / `Amount` `IsNil()` guards.

## Test plan
- [x] `go test ./x/stake/types/ -run NilAmount` (no panic; error contains `amount is required`)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/x/stake/types/msg.go
+++ b/x/stake/types/msg.go
@@ -70,6 +70,11 @@ func (msg MsgValidatorJoin) ValidateBasic() error {
 		return ErrInvalidMsg.Wrap("signer public key can't be of zero bytes")
 	}
 
+	// Guard before IsZero/IsNegative: a nil/uninitialized Int (proto field omitted)
+	// panics on those methods, and ValidateBasic runs ahead of signature verification.
+	if msg.Amount.IsNil() {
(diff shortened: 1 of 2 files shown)


SashaMIT:
Still relevant on current `develop`.

`MsgValidatorJoin.Amount` and `MsgStakeUpdate.NewAmount` still call `IsZero()` / `IsNegative()` without an `IsNil()` guard in `ValidateBasic`. An omitted proto amount still panics before signature verification.

Happy to rebase if you want this landed.

pratikspatil024:
Hi could you please rebase/merge latest develop?

SashaMIT:
Rebased onto current develop.

pratikspatil024 (approved):

